PR #66, opened 2014-06-24
BZ1032308 Warn if node address not found in rhq-storage-auth.conf

genman:
Warnings added, will stop after 5 times.

Fixed possible file descriptor leak.

Added a test to test the class and warning output.


rhqci:

PR #79, opened 2014-07-08
Bz1116914

genman:
When alerting for resource group, this change fixes notification. The UI is still not entirely functional.

The NPE in notify changes are mostly whitespace, there is very little logic change except for the email that is actually sent.


rhqci:

PR #110, opened 2014-08-14
BZ 1124969 - change ConfigEditor to only create backup before save

genman:
Also fix Cassandra plugin discovery to not rely on accessing the process
current working directory. This happens when the agent is run as a different
(non-root) user than the server. When this happens, the working directory
is inferred by the command line.


rhqci:

PR #124, opened 2014-09-08
BZ1101038 : Discover the <Location> Apache directive

jerr:
Lite pull request to discover <Location> Apache directive.
Availability check the URL.

Hope that is correct for you.
Regards,
Jérémie.


rhqci:

PR #133, opened 2014-10-05
Bz1103426 : Metrics for Apache server-status handler in <Location> directive

jerr:
It parses the status page to retrieve data from the scoreboard.
It depends on the Bz1101038 pull request.

Regards,
Jérémie.


rhqci:

maxime-beck:
The fix is OK.

As a side note, the default BMX Handler URL is pulled from the Listen value in the Apache configuration. If this value is set to ***:port**, mod_bmx values are not retrieved.

PR #142, opened 2014-10-17
Bug 1132720 - delete the 'illegitimate' plugins, not back them up

genman:
I also introduced a small change, where if sleep is interrupted (like during shutdown), it's not caught and cause a hang. Maybe it belongs in its own commit.


rhqci:

PR #153, opened 2014-12-12
Bug 1173096 - Support for Filtering child resources

lzoubek:
First commit adds:
/resource/{id}/children that could accept the same @QueryParams as /resource
- child resources can now be filtered by name status and category

2nd commit adds:
Added new endpoint : GET /resource/search which supports *ALL* filtering
options from ResourceCriteria class with small exception - it cannot pass
multiple values to filter functions.

2nd commit implementation has an advantage of being generic (if we happen to add a filter to ResourceCriteria, there is a fair chance we don't have to touch REST code, but also a disadvantage - documentation - every parameter could not be documented. I'd like to ask for review .. wheter we want to go this way. We may merge only first commit, but I think the 2nd one adds a real power to REST API to resource filtering


lzoubek:
retest this please

PR #157, opened 2015-02-10
Bug 1191063 - Tomcat plugin can't deploy packages to tomcat server by Bu...

Sapphire1:
...ndle

Add property deployDir and bundle-target.
	modified:   modules/plugins/tomcat/src/main/resources/META-INF/rhq-plugin.xml

### DIFF
--- a/modules/plugins/tomcat/src/main/resources/META-INF/rhq-plugin.xml
+++ b/modules/plugins/tomcat/src/main/resources/META-INF/rhq-plugin.xml
@@ -23,6 +23,11 @@
          <c:group
             name="connection"
             displayName="Connection Info">
+	    <c:simple-property name="deployDir"
+	       displayName="Deployment Directory" 
+	       type="directory" readOnly="false"    
+	       required="false"  
+	       description="The deployment directory path of the Tomcat server."/>
             <c:simple-property
                name="installationPath"
                displayName="Catalina Home"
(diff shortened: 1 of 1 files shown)


modules/plugins/tomcat/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -221,7 +226,26 @@
...
+	    <value-name>deployDir</value-name>
+	</destination-base-dir>
+	<destination-definition name="Patches">
+	    <accepts bundle-type="Tomcat Patch"/>
+	    <description>Use this location to apply patches to the server</description>
+	    <connection>WEBAPPS</connection>
+	    <referenced-configuration>
+		<all of="pluginConfiguration"/>
+	    </referenced-configuration>
+	</destination-definition>
```
lzoubek: What is the reason for <destination-definition name="Patches"> ? I am not aware of tomcat being capable of applying patches and also "Tomcat Patch" bundle type is not known to me. Could you please give a hint?

In case you just copy&pasted this stuff from AS7 plugin, please remove it.

Sapphire1:
Sorry, it was mindless copy&paste. I removed it.

lzoubek:
@Sapphire1 sorry for late response .. I was going through your PR again .. and I wonder if "deployDir" pluginConfiguration property could be discovered and maybe if it could be set as required (as it's value is required in bundle deployment)

rhqci:

PR #162, opened 2015-03-27
[BZ 1147098] Queue all the numeric data before writing to the StorageNode

burmanm:
This patch accepts data from the agents if it can write it to the in-memory queue. Otherwise the agent receives an exception and it should resend the data on later time. This is to balance the load on the RHQ Server if the storage node is incapable of accepting configured amounts of writes per second. It also tries to prevent a loss of data by returning it back to the agent for later processing.


jsanda:
Have you done any load/stress testing with these changes? I've used agentcopy with the perftest plugin in the past to generate large write loads.

burmanm:
What would you view as the target for the perf-testing? There are no changes to the actual Cassandra writes -> the performance bottleneck of the driver has not been changed at all.

But there's something weird about this PR, this is missing commits it seems.

jsanda:
The write path between RHQ and Cassandra has not changed, but the overall flow from source to sink certainly has.

burmanm:
The BlockingQueue part is much faster than what we're able to ingest or write to Cassandra if that's what you mean.

PR #163, opened 2015-03-30
Bug 1188751 - Deploying Versioned WAR package from repository to JBOSS EAP 6  deployment resource fails

lzoubek:
Support for updating package content for versioned deployments. This commit
also extracts code for creating deployment resources into Deployer class.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseComponent.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseComponent.java
@@ -23,7 +23,6 @@ import static org.rhq.modules.plugins.jbossas7.ASConnection.verbose;
 
 import java.io.OutputStream;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


jshaughn:
The approach defers to the existing discovery code but resolves the deployed package content name issues, so I think it is good.  Make sure to test with a versioned ear that contains a versioned war.  And also an upgrade of that ear.  Good job on enhancing this tricky feature.

lzoubek:
I need to add few more commits, to handle the same logic in DomainDeployment case. In that case, DomainDeployment may be assigned to serveral serverGroups. In that case, I need to unassign the original deployment from all assigned groups, then undeploy it, then deploy new content and then assign the new deployment to same serverGroups.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseComponent.java
```diff
@@ -544,64 +542,12 @@ protected CreateResourceReport deployContent(CreateResourceReport report) {
     public CreateResourceReport runDeploymentMagicOnServer(CreateResourceReport report, String runtimeName,
         String deploymentName, String hash) {
 
-        boolean toServerGroup = context.getResourceKey().contains("server-group=");
```
lzoubek: distinction between whether we're creating serverGroup deployment or not is no longer needed - serverGroups don't support adding Deployment child anymore

PR #166, opened 2015-04-22
[JWS-54] JON Tomcat8: Not discovered default web app config values

maximebeck:


rhqci:

lzoubek:
ok to test

PR #169, opened 2015-04-30
Bug 1212950 - EAP 6 start operation causes agent to run out of memory du...

lzoubek:
...e to

storing console output in an unused buffer

Now process output is captured (if captured) up to 2MB size, once output
exceeds this limit, it is ignored - so we don't run out of memory (unless
agent does not start plenty of verbose processes). Default limit can be
changed via rhq.process-execution.captured-output.limit system property.

This commit also gives more power to plugin writers about capturing process
outputs. ProcessExecution#setCaptureOutput is now deprecated in favor of new
CaptureMode setting. CaptureMode can capture to memory and/or forward to
agent.log as well as setting captured limit.

### DIFF
--- a/modules/core/native-system/src/main/java/org/rhq/core/system/ProcessExecutionOutputStream.java
+++ b/modules/core/native-system/src/main/java/org/rhq/core/system/ProcessExecutionOutputStream.java
@@ -1,0 +1,90 @@
+/*
+ * RHQ Management Platform
+ * Copyright (C) 2005-2015 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License, version 2, as
+ * published by the Free Software Foundation, and/or the GNU Lesser
+ * General Public License, version 2.1, also as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with this program;
+ * if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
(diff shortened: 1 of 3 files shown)


tsegismont:
@lzoubek thanks for the PR. I'll try to review it today.

modules/core/native-system/src/main/java/org/rhq/core/system/ProcessExecutionOutputStream.java
```diff
@@ -0,0 +1,103 @@
...
+    static {
+        int limit = 2 * 1024 * 1024; // default 2MB
+        try {
+            limit = Integer.parseInt(System.getProperty("rhq.process-execution.captured-output.limit",
+                String.valueOf(limit)));
+        } catch (NumberFormatException nfe) {
+        }
+        DEFAULT_LIMIT = limit;
+    }
+
```
tsegismont: I think a call to `java.lang.Integer#getInteger(java.lang.String, int)` would be easier to read (and does the same thing).
```java
private static final int DEFAULT_LIMIT = Integer.getInteger("rhq.process-execution.captured-output.limit", 2 * 1024 * 1024);  // default 2MB
```

modules/core/native-system/src/main/java/org/rhq/core/system/ProcessExecutionOutputStream.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public synchronized void setWriteToLog(boolean writeToLog) {
+        this.writeToLog = writeToLog;
+    }
+
+    /**
+     * sets new buffer limit
+     * @param limit in Bytes
+     */
+    public synchronized void setLimit(int limit) {
```
tsegismont: In my opinion, `writeToLog` and `limit` should be set when creating the instance and not be modifiable. It does not really make sense to change them once the instance is created.

tsegismont:
Other than the two comments inline, the change looks good to me. Could you replace calls to the deprecated method in the codebase? Or at least in the EAP 6 plugin ?

lzoubek:
@tsegismont thank you for reviewing, I've fixed and rebased my original commit. I'll replace deprecated calls in EAP6 plugin within a separate commit, I need to make sure it does not get to 3.3.x branch.

tsegismont:
OK. I understand the concern, but please make sure to mention that in the master commit otherwise if we get into any trouble then we'll spend some time before figuring out that part of the code is not in the 3.3.x branch.

PR #172, opened 2015-05-22
Bug 1224093 - AS7Based product plugins should not require changing

lzoubek:
JBossProductType when there is a change in slot value

JBossProductType is now deprecated in favor of JBossProduct class. New class
JBossProductDiscovery knows all JBossProducts (equivalent to previous JBoss
ProductType enums) and knows how to discover JBossProduct (reading
product.conf or fallback to detection by homeDir). Big change is, that if it
finds product.conf and it's slot value does not match any known
JBossProduct, it returns JBossProduct.unknown(). This value is to be
recognized by ResourceDiscoveryCallback implementations of derived plugins.
To make the life easier for product plugins based on AS7,
JBossProductDiscoveryCallback class has been added. Derived plugin can
extend it and only implement getProduct() method.


lzoubek:
retest this please

lzoubek:
retest this please

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/HostControllerDiscovery.java
```diff
@@ -65,24 +65,35 @@ protected String getLogFileName() {
     @Override
     protected String buildDefaultResourceName(HostPort hostPort, HostPort managementHostPort,
         JBossProductType productType, String serverName) {
+        return super.buildDefaultResourceName(hostPort, managementHostPort, productType, serverName);
```
tsegismont: Shouldn't this method be deprecated as well?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/HostControllerDiscovery.java
```diff
@@ -65,24 +65,35 @@ protected String getLogFileName() {
...
         }
 
-        return String.format("%s %s (%s:%d)", productType.SHORT_NAME, instanceType, managementHostPort.host,
+        return String.format("%s %s (%s:%d)", product.SHORT_NAME, instanceType, managementHostPort.host,
             managementHostPort.port);
     }
 
     @Override
     protected String buildDefaultResourceDescription(HostPort hostPort, JBossProductType productType) {
+        return super.buildDefaultResourceDescription(hostPort, productType);
```
tsegismont: Same here?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProduct.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    public static JBossProduct jdv() {
+        return new JBossProduct(JBossProductType.JDV, "dv", null);
+    }
+
+    public static JBossProduct unknown() {
+        return unknown(null);
+    }
+
```
tsegismont: Maybe we can cache all these (in a static final field) to avoid creating instances over and over?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/ServerPluginConfiguration.java
```diff
@@ -198,15 +201,35 @@ public void setLogDir(File logDir) {
         this.pluginConfig.setSimpleValue(Property.LOG_DIR, (logDir != null) ? logDir.toString() : null);
     }
 
+    @Deprecated
```
tsegismont: Can you add some Javadoc to indicate the version from which this method is deprecated? Same applies to other deprecated methods.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
```diff
@@ -82,17 +82,28 @@ protected String getLogFileName() {
...
     }
 
     @Override
     protected String buildDefaultResourceDescription(HostPort hostPort, JBossProductType productType) {
-        return String.format("Standalone %s server", productType.FULL_NAME);
+        return super.buildDefaultResourceDescription(hostPort, productType);
+    }
+
+    @Override
+    public String buildDefaultResourceDescription(HostPort hostPort, JBossProduct product) {
```
tsegismont: Do not augment visibility (keep it package protected like in the parent class), to avoid future problems with API compatibility.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
```diff
@@ -82,17 +82,28 @@ protected String getLogFileName() {
     @Override
     protected String buildDefaultResourceName(HostPort hostPort, HostPort managementHostPort,
         JBossProductType productType, String serverName) {
+        return super.buildDefaultResourceName(hostPort, managementHostPort, productType, serverName);
```
tsegismont: Deprecate?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
```diff
@@ -82,17 +82,28 @@ protected String getLogFileName() {
...
         }
 
-        return String.format("%s (%s:%d)", productType.SHORT_NAME, managementHostPort.host, managementHostPort.port);
+        return String.format("%s (%s:%d)", product.SHORT_NAME, managementHostPort.host, managementHostPort.port);
     }
 
     @Override
     protected String buildDefaultResourceDescription(HostPort hostPort, JBossProductType productType) {
-        return String.format("Standalone %s server", productType.FULL_NAME);
+        return super.buildDefaultResourceDescription(hostPort, productType);
```
tsegismont: Deprecate?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProduct.java
lzoubek: OK

tsegismont:
I wonder if the JBossProduct class should also indicate if the product supports patching. Then new products which we don't already know could set the supportsPatching attribute with the new discovery callback utility. What do you think?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
lzoubek: OK, I'll deprecate all you mentioned

tsegismont:
Can you update `org.rhq.modules.plugins.jbossas7.BaseServerComponent#validateServerAttributes` so that it no longer uses the hardcoded `expectedRuntimeProductName` plugin config property?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
lzoubek: I intentionally made this public so this 2 methods could be called from ResourceDiscoveryCallback implementations in other plugins. But at the same time I provided abstract callback implementation (in this PR), which should be enough for plugins. So it's public "just in case" plugin writer decides to implement his callback without extending JBossProductDiscoveryCallback and still wants to set resource name/description the general way. Do you think this use case is not likely? I am not sure

tsegismont: Le 28/05/2015 10:22, Libor Zoubek a écrit :
> In
> modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/StandaloneASDiscovery.java
> <https://github.com/rhq-project/rhq/pull/172#discussion_r31212171>:
>
>>      }
>>
>>      @Override
>>      protected String buildDefaultResourceDescription(HostPort hostPort, JBossProductType productType) {
>> -        return String.format("Standalone %s server", productType.FULL_NAME);
>> +        return super.buildDefaultResourceDescription(hostPort, productType);
>> +    }
>> +
>> +    @Override
>> +    public String buildDefaultResourceDescription(HostPort hostPort, JBossProduct product) {
>
> I intentionally made this public so this 2 methods could be called from
> ResourceDiscoveryCallback implementations in other plugins. But at the
> same time I provided abstract callback implementation (in this PR),
> which should be enough for plugins. So it's public "just in case" plugin
> writer decides to implement his callback without extending
> JBossProductDiscoveryCallback and still wants to set resource
> name/description the general way. Do you think this use case is not
> likely? I am not sure
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rhq-project/rhq/pull/172/files#r31212171>.
>


I'd rather keep it package-protected and tell plugin implementors to use 
JBossProductDiscoveryCallback in the future.

lzoubek:
I'll take a look, but I remember that supportsPatching detection makes decision also based on OSGI version.

Yes, I'll update ```expectedRuntimeProductName``` reference. I've "discovered" ```ServerPluginConfiguration.Property``` class when writing this patch and I saw lots of places, where we should be using it. I did not want this patch to refactor unrelated stuff, but for ```expectedRuntimeProductName``` it makes sense.

tsegismont:
> I'll take a look, but I remember that supportsPatching detection makes decision also based on OSGI version.

Yes that would imply to have version info in the jboss product type class as well. If you feel this is going to put things at risk (too many impacts), don't do it. I'm just wondering if it's easily doable.

> Yes, I'll update expectedRuntimeProductName reference. I've "discovered" ServerPluginConfiguration.Property class when writing this patch and I saw lots of places, where we should be using it. I did not want this patch to refactor unrelated stuff, but for expectedRuntimeProductName it makes sense.

Cool. There are other places in the code where the hardcoded version is used, I think.

tsegismont:
Have you tried to test your build with ISPN and the ISPN plugin. That would be a good indication that the changes are working properly.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscoveryCallback.java
```diff
@@ -0,0 +1,159 @@
...
+import org.rhq.modules.plugins.jbossas7.helper.HostPort;
+import org.rhq.modules.plugins.jbossas7.helper.ServerPluginConfiguration;
+
+/**
+ * Default abstract implementation of {@link ResourceDiscoveryCallback} interface, that may be extended by
+ * AS7 based plugins. This implementation handles cases when discovered JBossProduct equals to {@link JBossProduct#unknown()} - this means JBossAS7 plugin
+ * does know definition of JBossProduct it detected. The extending class only needs to implement {@link #getProduct()}.
+ * @author lzoubek@redhat.com
+ *
+ */
```
tsegismont: Do you expect different usages of this class? If we only expect implementors to override `getProduct` , then let's not expose the rest of the class, and document here in the class Javadoc what's the class behavior.

tsegismont:
Is it possible to test the discovery callback utility? With a fake plugin and a fake product?

tsegismont:
Overall, very nice job Libor! Waiting for your answers/updates before merging.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscoveryCallback.java
lzoubek: On Thu, 28 May 2015 10:58:19 +0200, Thomas Segismont  
<notifications@github.com> wrote:

> Do you expect different usages of this class? If we only expect  
> implementors to override `getProduct` , then let's not expose the rest  
> of the class, and document here in the class Javadoc what's the class  
> behavior.
>
```getProduct``` is just the minimum. I also expect known product (like  
ISPN,JDV) plugins, could us this as base class, once they need (at least  
it gives them a bit more AS7 specific context). Thatswhy I did most of the  
stuff as protected. For example if gatein discovery callback is rewritten  
it can consume ```ServerPluginConfiguration.setExpectedRuntimeName```  
instead of (right now) hardcoded ```expectedRuntimeName``` string.
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/rhq-project/rhq/pull/172/files#r31214536


-- 
Libor Zoubek

**Comment by lzoubek** [comment]
On Thu, 28 May 2015 11:08:56 +0200, Thomas Segismont  
<notifications@github.com> wrote:

> Is it possible to test the discovery callback utility? With a fake  
> plugin and a fake product?
That is quite hard. I was testing it with RHQ Server (I did local changes  
to rhqserver plugin), discoveryCallback worked fine, but there was an  
issue with server itself (because I changed it's slot from ```eap``` to  
```jon``) I was not able to respond on http management interface. I  
suspect this is due to my lack of knowledge about what slot really means  
in EAP.

If you have an idea how to automate this, I'll be happy to do that.
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/rhq-project/rhq/pull/172#issuecomment-106242447


-- 
Libor Zoubek

**Comment by lzoubek** [comment]
On Thu, 28 May 2015 10:48:37 +0200, Thomas Segismont  
<notifications@github.com> wrote:

> Have you tried to test your build with ISPN and the ISPN plugin. That  
> would be a good indication that the changes are working properly.
Great idea, will do.

**Comment by tsegismont** [comment]
Le 28/05/2015 11:59, Libor Zoubek a écrit :
>
> If you have an idea how to automate this, I'll be happy to do that.

No I don't, but manual testing with ISPN and its plugin is a first step

**Comment by tsegismont** [comment]
Le 28/05/2015 11:59, Libor Zoubek a écrit :
> Great idea, will do.

Perfect, thanks

File: modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscoveryCallback.java (hunk 8 above)
Comment (tsegismont) [inline modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscoveryCallback.java]: Ok

File: modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscovery.java [hunk 9]
```diff
@@ -0,0 +1,162 @@
...
+        Iterator<JBossProduct> products = Arrays.asList(
+            JBossProduct.AS,
+            JBossProduct.EAP,
+            JBossProduct.WILDFLY8,
+            JBossProduct.ISPN,
+            JBossProduct.JDG,
+            JBossProduct.JPP,
+            JBossProduct.SOA,
+            JBossProduct.JDV,
+            JBossProduct.unknown()
```
tsegismont: Why exactly do we need  JBossProduct.unknown() to be in this list?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscovery.java
```diff
@@ -0,0 +1,162 @@
...
+        return product;
+    }
+
+    public static JBossProduct getKnownProductByProductName(String productName) {
+        for (JBossProduct product : KNOWN_PRODUCTS.values()) {
+            if (product.PRODUCT_NAME.equals(productName)) {
+                return product;
+            }
+        }
+        LOG.debug("No product type with product-name '" + productName + "' is known. returning UNKNOWN");
```
tsegismont: Make sure debug logs with string concats are wrapped wih if LOG.isDebugEnabled

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscovery.java
lzoubek: I left it in list because of ```JBossProductDiscovery#getKnownProduct*``` methods to suppress debug messages in case "UNKNOWN" product type is requested. Those methods will produce log message in case given product is not found among KNOWN_PRODUCTS. JBossProduct.unknown() can be persisted into serverPluginConfiguration for some time (before discoveryCallback of particular plugin runs). In theory, AS7 plugin can discover an unknown product, and after the right plugin with discoveryCallback is installed, it will finish the product discovery.

I can remove it if you don't like it. Then I'll fix those 2 methods

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscovery.java
lzoubek: ok

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductDiscovery.java
tsegismont: Are you sure this how discovery callbacks works? My understanding was callbacks were called after the plugin container got the PluginConfig details from AS7 plugin.

PR #173, opened 2015-06-01
Bug 1220277 - Correctly handle unsufficient permission errors

lzoubek:
This fixed BaseServerComponent only - so we display correct resource error,
when insufficient permissions are detected. In all other cases (create
child, operations) it should be obvious from EAP's response which appears on
top of stacktrace. This commit also fixes an issue when there are invalid
user/pass set in pluginConfiguration resource error was not reported
(because InvalidPluginConfigurationException thrown by AS7Connection.execute
was swallowed by getAvailability() code.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseServerComponent.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseServerComponent.java
@@ -79,6 +79,7 @@ import org.rhq.modules.plugins.jbossas7.json.ReadResource;
 import org.rhq.modules.plugins.jbossas7.json.Result;
 import org.rhq.modules.plugins.jbossas7.json.ResultFailedException;
 import org.rhq.modules.plugins.jbossas7.json.SecurityRealmNotReadyException;
+import org.rhq.modules.plugins.jbossas7.json.UnauthorizedException;
 
 /**
  * Base component for functionality that is common to Standalone Servers and Host Controllers.
@@ -128,6 +129,7 @@ public abstract class BaseServerComponent<T extends ResourceComponent<?>> extend
     public AvailabilityType getAvailability() {
         AvailabilityType availabilityType;
         boolean securityRealmReady = true;
+        boolean accessPermission = true;
         try {
             try {
                 readAttribute(getHostAddress(), "name", AVAIL_OP_TIMEOUT_SECONDS);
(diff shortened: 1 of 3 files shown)


modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseServerComponent.java
```diff
@@ -143,6 +144,12 @@ public AvailabilityType getAvailability() {
                 previousAvailabilityType = DOWN;
                 availabilityType = DOWN;
                 securityRealmReady = false;
+            } catch (UnauthorizedException e) {
+                previousAvailabilityType = DOWN;
+                availabilityType = DOWN;
+                throw new InvalidPluginConfigurationException("User authenticated via HTTP management interface does"
```
tsegismont: Instead of throwing the exception here, you could set a boolean (see `securityRealmReady` above). Then you throw the IPCE later, which avoids to catch/rethrow

tsegismont:
Apart for the one comment inline, looks good to me. Thanks Libor!

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseServerComponent.java
lzoubek: I followed ```securityRealmReady``` approach first, but then I decided do to with rethrowing. Catch block rethrowing IPCE is needed anyway, because ```ASConnection.execute()``` throws it  when credentials are incorrect. I'll fix the code as you suggest and avoid rethrowing

tsegismont: Le 02/06/2015 10:47, Libor Zoubek a écrit :
> Catch block rethrowing IPCE is needed anyway, because
> |ASConnection.execute()| throws it when credentials are incorrect

Mmm, I don't think it's needed. If ASConnection#execute throws IPCE 
because of invalid credentials then we should just let the exception 
bubble up.

lzoubek: >
> Mmm, I don't think it's needed. If ASConnection#execute throws IPCE
> because of invalid credentials then we should just let the exception
> bubble up.

It won't bubble up, because it's caught by  
https://github.com/lzoubek/rhq/blob/bugs/1220277/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseServerComponent.java#L169  
Then resource error is not stored and user does not know why's the  
resource down (this should be separate BZ)

tsegismont: Right, sorry for missing that.

PR #176, opened 2015-06-09
Bug 1227459 - Allow using $local authentication method when executing…

lzoubek:
… CLI

scripts

Added new pluginConfig boolean property "Native Local Authentication". This
gets discovered to true only if $local authentication is the only way to
access native interface.

### DIFF
--- a/modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/XmlFileReadingTest.java
+++ b/modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/XmlFileReadingTest.java
@@ -153,4 +153,18 @@ public class XmlFileReadingTest {
         assert "jboss.server.config.dir".equals(propsFilePathRel) : "Path was " + propsFileName;
     }
 
+    public void testIsLocalNativeAuth() throws Exception {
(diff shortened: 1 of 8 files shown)


modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/XmlFileReadingTest.java
```diff
@@ -153,4 +153,18 @@ public void testGetRealm() throws Exception {
         assert "jboss.server.config.dir".equals(propsFilePathRel) : "Path was " + propsFileName;
     }
 
+    public void testIsLocalNativeAuth() throws Exception {
```
tsegismont: Thanks for making this tested

tsegismont:
This looks good to me. I think we should add some resource upgrade code as well. So that users don't have to go through a manual resource pluginconfig update process.

tsegismont:
So you think resource upgrade is not a good idea?

lzoubek:
I think it is a good idea. I squashed the commits before merge. Thank you for reviewing this.

tsegismont:
Ok, it flew below my radar :smile:

PR #177, opened 2015-06-19
Bug 1226413 - Implementation of rhq:handover doesn't account for EAP https connections

lzoubek:
Anytime plugin attempts to run CLI in connected mode, we execute  dummy CLI
command to detect whether SSL certificate needs to be accepted (manual
intervention needed). This dummy command is executed prior the original
command so we can make sure the original command is not blocked forever by
CLI asking for user input. In case we detect (based on dummy command
output), we return special implementation of ProcessExecutionResults to
caller which holds results from dummy process with error message produced by
CLI representing certificate details.

This commit does not fix 100% Bug 1226413, but  prevents blocked CLI process
as part of rhq:handover action or CLI operation.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/ServerControl.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/ServerControl.java
@@ -203,6 +203,10 @@ public final class ServerControl {
 
     public final class Cli {
         private boolean disconnected;
+        /**
+         * whether to execute dummy CLI check prior running desired command
+         */
+        private boolean checkCertificate = true;
 
         Cli() {
             // When running the CLI on Windows, make sure no "pause" message is shown after script execution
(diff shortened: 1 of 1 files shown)


tsegismont:
I guess this is the only way to work around the issue. But this is ugly (not your fault). Can you check with EAP team if:
* there is a way to put JBoss CLI in non-interactive mode only
* or there's a property to make the certificate check fail if it's never been used before
* or there's an option to make JBoss CLI use the same truststore file as the RHQ agent?

lzoubek:
I'll ask EAP guys what you suggest.

There is a way to make JBoss CLI use the same truststore file as RHQ agent. This can be done by editing jboss-cli.xml (default configuration file for CLI) where we can write path to truststore + password. I am not sure if it is better way, since we either need to put the password in plain text, or using vault. I can imagine we can write to jboss-cli.xml, but not sure if we're able to create vault and store truststore password in it.

tsegismont:
Looking at the schema, there's a "modify-trust-store" option in ssl 
config which can switch off the accept certificate prompt

lzoubek:
It is, but it does not switch the prompt off. Unfortunately this only removes the option to accept the certificate permanently.

tsegismont:
Doh!

PR #178, opened 2015-06-23
[BZ 1212627] Recreate storage node sessions if connections are down

burmanm:
Check storage node connection aliveness every 4s and recreate the session if check failed twice in a row. This causes the RHQ server to be able to reconnect to the storage nodes in case the whole cluster went down.

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
@@ -191,6 +191,7 @@ public class StorageClientManager implements StorageClientManagerMBean{
             initMetricsServer();
             JMXUtil.registerMBean(this, OBJECT_NAME);
             aliveChecker = new SessionAliveChecker(this);
+            aliveChecker.setName("StorageNode SessionAliveChecker");
             aliveChecker.start();
 
             initialized = true;
@@ -688,6 +689,8 @@ public class StorageClientManager implements StorageClientManagerMBean{
                     }
                 } catch (InterruptedException e) {
                     // Alive should be false in the next iteration if shutdown was called
+                } catch(Exception e) {
+                    LOG.error("AliveCheck thread run into an unexpected exception: " + e.getLocalizedMessage());
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
```diff
@@ -236,26 +240,36 @@ public synchronized boolean refreshCredentialsAndSession() {
...
-                    + " as soon as possible.", e);
-                return false;
+    /**
+     * Recreates the session used to connect to the storage node.
+     * @return true if success, false otherwise.
+     */
+    private synchronized boolean refreshSession() {
+        Session wrappedSession;
+        try {
+            wrappedSession = createSession();
```
jsanda: We should shut down the old session; otherwise, we will be leaking connections and threads.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
```diff
@@ -631,4 +647,54 @@ public int getQueueAvailableCapacity() {
...
+                        if(fails >= ALLOWED_FAILS) {
+                            LOG.error("Failed to contact the storage node for live check, recreating connection session");
+                            // We have lost the connection to the storage node, refresh and try again..
+                            storageClientManager.refreshSession();
+                            Thread.sleep(EXTENDED_SLEEP); // Sleep for a longer time to allow storage node to restart
+                        }
+                    }
+                } catch (InterruptedException e) {
+                    // Alive should be false in the next iteration if shutdown was called
+                }
```
jsanda: I would be inclined to add another catch block here for Exception or even possibly Throwable. If a RuntimeException is thrown for example, the SessionAliveChecker thread will die silently.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
burmanm: Old session is already shutdown in StorageSession.registerNewSession(Session)

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageClientManager.java
burmanm: Yea, will add. I actually found a bug while missing this addition (the rate was set to 0.0 when storage nodes went down and InvalidArgumentException was thrown by the Guava)

jsanda:
I missed the shutdown call in registerNewSession(). LGTM.

PR #179, opened 2015-06-30
Bug 1236631 - Add a way to setup SSL settings in jboss-cli.xml

lzoubek:
Added operation called "Setup CLI" which can change jboss-cli.xml according
to pluginConfiguration properties. It can configure SSL stuff + default
controller host + port. Operation is present for Standalone and Host
Controllers.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/JBossCliConfiguration.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/JBossCliConfiguration.java
@@ -1,0 +1,389 @@
+/*
+ * RHQ Management Platform
+ * Copyright (C) 2005-2015 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+package org.rhq.modules.plugins.jbossas7.helper;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import javax.xml.namespace.QName;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.w3c.dom.Comment;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+import org.rhq.core.util.file.FileUtil;
+
+/**
+ * A JBoss CLI configuration - loaded from jboss-cli.xml
+ *
+ * @author Libor Zoubek
+ */
+public class JBossCliConfiguration {
+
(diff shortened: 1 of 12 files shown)


tsegismont:
Overall, looks good to me. Thanks for the documentation of the operation in the plugin descriptor.

I suppose having the truststore and keystore passwords in plain text is the only way to configure the CLI?

lzoubek:
Thank you for looking at this PR. Vaults can be used to store those. But creating a vault requires just another java truststore file. People could already be using vaults. Such operation would have to take path to truststore + password as inputs to store new vault. Few days I realized it's not worthy doing, now I think it might not be a bad idea.

lzoubek:
I explored vault more, and it is going to be quite easy to setup jboss-cli.xml using vault. This will require a vault already being configured in standalone.xml (I won't implement something to setup/initialize new vault). Our plugin would read vault options and password information from standalone.xml and put them into jboss-cli.xml

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/JBossCliConfiguration.java
```diff
@@ -0,0 +1,243 @@
...
+        for (Entry<String, String> vaultOpt : vaultOptions.entrySet()) {
+            Element opt = addChildElement(vaultNode, "vault-option");
+            opt.setAttribute("name", vaultOpt.getKey());
+            opt.setAttribute("value", vaultOpt.getValue());
+        }
+
+        addChildElement(sslNode, "trust-store", serverIdentity.alias);
+        addChildElement(sslNode, "trust-store", serverIdentity.path);
+        // in standalone.xml vault value is referred as ${VAULT:...} we need to strip it for jboss-cli.xml
+        addChildElement(sslNode, "trust-store-password", stripBrackets(serverIdentity.keystorePassword));
```
tsegismont: Why is it inside brackets? Is it safe to simply strip them?

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/TruststoreConfig.java
```diff
@@ -0,0 +1,50 @@
...
+package org.rhq.modules.plugins.jbossas7.helper;
+
+import org.w3c.dom.Node;
+
+/**
+ * 
+ * @author lzoubek
+ *
+ */
+public class TruststoreConfig {
```
tsegismont: Not very object oriented. Could you make the fields private final, use a private constructor and add getters?

tsegismont:
Libor, for the next PRs, could you wait for the final review before squashing commits? It's easier to review additional commits.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/JBossCliConfiguration.java
lzoubek: IMHO it is. According to [docs](https://developer.jboss.org/wiki/JBossAS7SecuringPasswords), vaults have to be enclosed in ${} in standalone.xml file, and [here](https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.3/html-single/Security_Guide/index.html) must not be in jboss-cli.xml file. 

My testing also confirms this. 

I would ignore a potential case when someone can store password into system property, refer to it using ${} and then run this new operation to setup jboss-cli.xml using vault.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/TruststoreConfig.java
lzoubek: OK

lzoubek:
While I was searching through docs I found out, that jboss-cli.xml is supported since EAP 6.2 and vaulted passwords in jboss-cli.xml since EAP 6.3. 

For < 6.2 this operation will fail, because jboss-cli.xml won't exist. for 6.2 it may produce invalid content in case user tries to store passwords using vault. I'll need to add a detection of xml namespace

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/JBossCliConfiguration.java
tsegismont: OK, thanks for the details.

tsegismont:
> While I was searching through docs I found out, that jboss-cli.xml is supported since EAP 6.2 and vaulted passwords in jboss-cli.xml since EAP 6.3.

> For < 6.2 this operation will fail, because jboss-cli.xml won't exist. for 6.2 it may produce invalid content in case user tries to store passwords using vault. I'll need to add a detection of xml namespace

Nice catch! Please also update the documentation of the operation in the plugin descriptor.

tsegismont:
Have you tested the operation on Windows?

lzoubek:
@tsegismont I've pushed new commit taking care of changes in jboss-cli.xml versions. Testing on widnows is the most painful piece of this work, which I am going to do today

tsegismont:
> Testing on widnows is the most painful piece of this work, which I am going to do today
Good luck!

lzoubek:
retest this please

tsegismont:
LGTM. It can be merged when Windows testing is completed.

lzoubek:
retest this please

lzoubek:
windows test passed

PR #180, opened 2015-07-02
Bug 1226413 - Bundle rhq:handover can hang when management interface is

lzoubek:
using SSL and jboss-cli does not yet trust certs

Fixed potential OOM caused by EAP 6.4.x which is constantly filling stdout
when waiting for user to accept ssl cert.

Fixed cleaning up java process (executed dummy CLI command), which could
survive, it's parent because it got detached from it.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/ServerControl.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/ServerControl.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/ServerControl.java
```diff
@@ -236,38 +246,47 @@ public Cli disconnected(boolean disconnected) {
...
+            String rhqPid = UUID.randomUUID().toString();
             ProcessExecutionResults result = ServerControl.onServer(pluginConfiguration, serverMode, systemInfo)
                 .killingOnTimeout(true)
                 .waitingFor(10 * 1000L)
+                .ignoreOutput()// avoid potential OOM https://bugzilla.redhat.com/show_bug.cgi?id=1238263
                 .cli()
                     .disconnected(false)
                     .dontCheckCertificate() // avoid stack overflow
-                    .executeCliCommand(":whoami");
+                .executeCliCommand(":whoami", "-Drhq.as7.plugin.exec-id=" + rhqPid);
```
tsegismont: Please re-indent

tsegismont:
Perfect Libor, the changes look good to me. Can you make sure a ticket is open on the Wildfly team side so that, in the future, the CLI has a true non-interactive mode? When it happens we'll be able to switch off this code for recent EAP versions.

lzoubek:
Thank you for review, here's the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1238609

I am doing some testing on windows as well and I'll merge this myself then

PR #183, opened 2015-09-16
Bug 1232847 - Increase timeout for repair operation

lzoubek:
Increase timeout for all long running storage node operations to 7 days.

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageNodeOperationsHandlerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/storage/StorageNodeOperationsHandlerBean.java
@@ -15,7 +15,7 @@ import javax.persistence.PersistenceContext;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.joda.time.Hours;
+import org.joda.time.Days;
 
 import org.rhq.cassandra.schema.Table;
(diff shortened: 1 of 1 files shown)


jsanda:
Check with Larry and see what he thinks about bumping up the timeout to 7 days. I'd like to get his input on this.

lzoubek:
@loleary what do you think about bumping up the timeout to 7 days for potentially long running operations such as repair or addNodeMaintenance (which could include repair)

loleary:
7 days is fine. Of course I am not sure how easy it will be to test this. If memory serves me right, there is some hard-coded 10 hour or 24 hour timeout for operations somewhere so I fear that this change won't do what we need it to. But I really don't know for certain.

lzoubek:
[OperationManager](https://github.com/rhq-project/rhq/blob/master/modules/core/plugin-container/src/main/java/org/rhq/core/pc/operation/OperationManager.java#L219) defaults to 600s in case timeout is not defined within input parameters (in our case it is) or operation definition. I did not find anything about 10 or 24 hours.

lzoubek:
retest this please

lzoubek:
@loleary you we're right, I just tried to run an operation (which would end up after 25 hours) and it finished within ~24hours with weird result ```Failed to start : operation never started``` I'll take look at operations deeper

lzoubek:
Bug 1265309 (preventing operations running more than 24h) fixed by 1254b9fca merging this one

PR #184, opened 2015-09-16
Bug 1234912 - Do not authenticate against new storage node when

lzoubek:
replication_factor of system_auth keyspace is wrong

For system_auth keyspace set replication_factor=clusterSize, so each node
keeps it's own copy of auth data. Created recurrent job which checks
replication_factor for rhq and system_auth keyspacesm when invalid
replication_factor is detected, job tries to fix it and then recommends
running clusterMaintenance

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/scheduler/jobs/ReplicationFactorCheckJob.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/scheduler/jobs/ReplicationFactorCheckJob.java
@@ -1,0 +1,230 @@
+/*
+ * RHQ Management Platform
+ * Copyright (C) 2005-2015 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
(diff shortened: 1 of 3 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/scheduler/jobs/ReplicationFactorCheckJob.java
```diff
@@ -0,0 +1,230 @@
...
+                    inconsistency += "keyspace [" + factorEntry.getKey() + "] has replication_factor=" + current
+                        + " but it should be " + healthy + ", ";
+                    factorsToSet.put(factorEntry.getKey(), healthy);
+                }
+            }
+
+            if (!factorsToSet.isEmpty()) {
+                // strip last ", "
+                inconsistency = inconsistency.substring(0, inconsistency.length() - 2);
+                log.error("Storage Cluster is not consistent! There are "
```
jsanda: I might reduce the log level to WARN since this is a situation which we can handle or at least try to handle. In the log statement in addition to the number of storage nodes in the RDBMS, it would be helpful to also include the current RF and the expected RF.

Lastly, there is misspelling in rational.

lzoubek: OK, I thought ERROR level would be better because of importance of such state - WARNING gets usually ignored (including me). current and expected RFs are included in ```incosistency``` string, because expected RF can differ for rhq and system_auth keyspaces.

jsanda: >WARNING gets usually ignored (including me)

That is a good point :)

lzoubek:
@jsanda fixed (rebased)

jsanda:
The changes look good, but there is one thing you might want to consider adding. When the replication_factor of the system keyspace is wrong, particularly going from one to two nodes, that is when users most often run into the problems we are trying to address. Let's say the first storage node is down. The server will not be able to authenticate against the second (new) storage node because it does not yet have the authentication data. I think we want to run repair on the system_auth keyspace as soon as possible to avoid problems like that. Repair on the system_auth keyspace should be fairly fast as the amount of data stored in it is very small; however, as in my example in which the first node is down, running repair would be pointless.

lzoubek:
I'll open new BZ for adding repair capability to `ReplicationFactorCheckJob`

PR #185, opened 2015-09-21
Bug 1230411 - Storage node results in 1000s of configuration changes …

lzoubek:
…filling

up the database

Several Configuration properties were transformed to metrics or operations.

StorageService resourceType: LoadMap and Ownership config properties were
removed (metrics already existed) we're "loosing" information about other
nodes (which can be obtained on those node resources). TokenToEndpointMap
property was removed and transformed to operation called "View Token To Map
Ednpoint"

MessagingService resourceType:DroppedMessages and RecentlyDroppedMessages
transformed to operations, all other config properties transformed to
metrics (no properties left)

### DIFF
--- a/modules/plugins/cassandra/src/main/java/org/rhq/plugins/cassandra/MessagingServiceComponent.java
+++ b/modules/plugins/cassandra/src/main/java/org/rhq/plugins/cassandra/MessagingServiceComponent.java
@@ -1,0 +1,43 @@
+/*
+ *
+ *  * RHQ Management Platform
+ *  * Copyright (C) 2005-2015 Red Hat, Inc.
+ *  * All rights reserved.
+ *  *
+ *  * This program is free software; you can redistribute it and/or modify
+ *  * it under the terms of the GNU General Public License, version 2, as
+ *  * published by the Free Software Foundation, and/or the GNU Lesser
+ *  * General Public License, version 2.1, also as published by the Free
+ *  * Software Foundation.
+ *  *
+ *  * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


modules/plugins/cassandra/src/main/java/org/rhq/plugins/cassandra/MessagingServiceComponent.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * 
+ * @author lzoubek
+ *
+ */
+public class MessagingServiceComponent extends ComplexConfigurationResourceComponent {
+
+    public OperationResult invokeOperation(String name, Configuration parameters) throws Exception {
+        if ("DroppedMessages".equals(name) || "RecentlyDroppedMessages".equals(name)) {
```
jsanda: Why were these made operations and not metrics?

jsanda:
Other than my question about MessagingServiceComponent, I think the changes look good.

modules/plugins/cassandra/src/main/java/org/rhq/plugins/cassandra/MessagingServiceComponent.java
lzoubek: I'd have to introduce 16 new metrics, and still I was not sure whether this set is dynamic or not. Having 2 operations was easier.

jsanda: Cool. Thanks for the explanation!

PR #187, opened 2015-10-24
Bug 875156 - NPE in code completion

josejulio:
Method getContextMatches from JavascriptCompletor now verifies if the baseObject is not null.


rhqci:

PR #188, opened 2015-10-24
Bug 907897 - pretty.print(null) fails

josejulio:
Renamed the print overloaded methods to only leave one and avoid ambiguities.
void print(Object)
void printCollection(Collection)
void printConfiguration(Configuration)
...
The print method also verify the input to call the correct method.


rhqci:

PR #189, opened 2015-10-26
Bug 1155104 - CLI auto-complete works incorrectly on second . notation

josejulio:
Defered the check for "(" to allow chaining methods.
"completeParameters" no longer returns the params on the list (to avoid repetitions).


rhqci:

burmanm:
Hi, thanks for these fixes, I'll take a look at them a little later (this week).

josejulio:
Thanks.

PR #190, opened 2015-10-26
Bug 753204 - Discovery --full should be the default

josejulio:
Swapped the default discovery behavior with a --dry-run option and the  --full option.
(The new default is --full).

### DIFF
--- a/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/i18n/AgentI18NResourceKeys.java
+++ b/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/i18n/AgentI18NResourceKeys.java
@@ -1907,9 +1907,10 @@ public interface AgentI18NResourceKeys {
     @I18NMessage("discovery")
     String DISCOVERY = "PromptCommand.discovery";
 
-    @I18NMessage("discovery [--plugin=<plugin name>] [--resourceType=<type name>]\\n\\\n" //
+    @I18NMessage("discovery [--verbose]\\n\\\n"
+        + "\\        discovery --dry-run [--plugin=<plugin name>]\\n\\\n"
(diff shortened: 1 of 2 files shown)


rhqci:

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/i18n/AgentI18NResourceKeys.java
```diff
@@ -1917,15 +1918,13 @@
...
+        + "resources a plugin can actually find.  Note that this will run a detailed server\\n\\\n "
+        + "scan inside the plugin container.\\n\\\n"
+        + "This will update the actual agent inventory by sending an inventory report to the server.\\n\\\n"
+        + "Ignores --plugin, --resourceType and --resourceId unless --dry-run is used.\\n\\\n"
         + "The valid command line arguments are:\\n\\\n"
-        + "\\  -f, --full : Runs a detailed discovery inside the plugin container.\\n\\\n"
-        + "\\               This will update the actual agent inventory by sending\\n\\\n"
-        + "\\               an inventory report to the server.\\n\\\n"
-        + "\\               This ignores --plugin, --resourceType and --resourceId.\\n\\\n"
+        + "\\  -d, --dry-run : Runs a service scan (i.e. it will not try to discover child services\\n\\\n"
```
pilhuhn: Should that read "server scan" ?

pilhuhn:
Lgtm - @jmazzitelli can you have a look too?

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/i18n/AgentI18NResourceKeys.java
josejulio: Yes. My mistake, I'll fix it.

PR #191, opened 2015-10-26
Bug 833547 - make agent shutdown timeout configurable

josejulio:
Takes the configuration from the AgentConfiguration.
Created the defaults in the AgentConfigurationConstants.
The configuration is taken from the entry: "rhq.agent.agent-update-exit-timeout-msecs".

### DIFF
--- a/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfiguration.java
+++ b/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfiguration.java
@@ -27,8 +27,6 @@ import java.util.List;
 import java.util.prefs.BackingStoreException;
 import java.util.prefs.Preferences;
 
-import mazz.i18n.Logger;
-
 import org.rhq.core.pc.PluginContainerConfiguration;
 import org.rhq.core.util.obfuscation.ObfuscatedPreferences;
 import org.rhq.enterprise.agent.i18n.AgentI18NFactory;
@@ -37,6 +35,8 @@ import org.rhq.enterprise.communications.ServiceContainerConfiguration;
 import org.rhq.enterprise.communications.command.client.ClientCommandSenderConfiguration;
 import org.rhq.enterprise.communications.command.client.PersistentFifo;
 import org.rhq.enterprise.communications.util.SecurityUtil;
+
+import mazz.i18n.Logger;
 
 /**
  * Just provides some convienence methods to extract agent configuration properties.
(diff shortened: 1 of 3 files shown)


rhqci:

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfigurationConstants.java
```diff
@@ -162,6 +162,16 @@
     String AGENT_UPDATE_DOWNLOAD_URL = PROPERTY_NAME_PREFIX + "agent-update.download-url";
 
     /**
+     * The amount of milliseconds the agent will wait before force closing after an update.
+     */
+    String AGENT_UPDATE_EXIT_TIMEOUT_MSECS = PROPERTY_NAME_PREFIX + "agent-update-exit-timeout-msecs";
+
```
josejulio: Not sure if I should change the name to: "AGENT_EXIT_TIMEOUT_MSECS", to cover as a global exit timeout (not sure if is or will be used anywhere else).

pilhuhn:
@jmazzitelli Can you please have a look?

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfiguration.java
```diff
@@ -374,6 +374,12 @@ public String getAgentUpdateVersionUrlIfDefined() {
         return str;
     }
 
+    public long getAgentUpdateExitTimeout() {
+        long value = m_preferences.getLong(AgentConfigurationConstants.AGENT_UPDATE_EXIT_TIMEOUT_MSECS,
```
burmanm: value is redundant, just return right away.

josejulio: I was just following the same notations that are along the file, should I just return right away then?

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfigurationConstants.java
jmazzitelli: name of the field is fine, but the value should be "agent-update.exit-timeout-msecs" - notice the "." after agent-update. The way these are define is we have the name of a group of properties followed by "." followed by the name of the property. Notice we have several agent-update properties - you are adding another:

agent-update.enabled
agent-update.version-url
agent-update.download-url
agent-update.exit-timeout-msecs <-- you want a "." after agent-update, like all the rest in the group.

josejulio: Right, missed it. Will fix it up, thanks.

modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentConfiguration.java
```diff
@@ -38,6 +36,8 @@
 import org.rhq.enterprise.communications.command.client.PersistentFifo;
 import org.rhq.enterprise.communications.util.SecurityUtil;
 
+import mazz.i18n.Logger;
```
josejulio: Eclipse moved this import, is OK?

PR #192, opened 2015-10-28
Fix JWS-118

gottburgm:
It was stopping at the first match.


rhqci:

PR #193, opened 2015-12-02
Promoted groovy-maven-plugin from -SNAPSHOT to non-snapshot to allow …

josejulio:
…building with -Pdev.


rhqci:

stefannegrea:
@josejulio, thanks for the PR.

This PR is needed because the -SNAPSHOT variant of that particular plugin is now missing from maven repositories.

PR #195, opened 2015-12-11
Bug 1268329 - The same user is able to upload bundle via 'Upload' but…

josejulio:
… not via 'URL'

Instead of returning the URL as the token, moved to the temp dir to match the special token handling worflow.
Unpacked the BundleNotFoundException from the RunTimeException that is throw to the client.
Was using BundleManager.createBundleVersionViaURL instead of BundleManager.createOrStoreBundleVersionViaURL (The caller method is "createOrStoreBundleVersionViaURL"), changed to match method signature.

### DIFF
--- a/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/bundle/create/BundleUploadDistroFileStep.java
+++ b/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/bundle/create/BundleUploadDistroFileStep.java
@@ -75,6 +75,9 @@ public class BundleUploadDistroFileStep extends AbstractWizardStep {
     private static final String URL_OPTION_PASSWORD = MSG.common_title_password();
     private static final String URL_OPTION_TOOLTIP = MSG.view_bundle_createWizard_urlTooltip();
 
+    final String BUNDLE_NOT_FOUND_EXCEPTION_PATTERN_START = "org.rhq.core.domain.bundle.BundleNotFoundException:[";
+    final String BUNDLE_NOT_FOUND_EXCEPTION_PATTERN_END = "]";
+
     public BundleUploadDistroFileStep(AbstractBundleCreateWizard bundleCreationWizard) {
         this.wizard = bundleCreationWizard;
     }
@@ -282,8 +285,9 @@ public class BundleUploadDistroFileStep extends AbstractWizardStep {
                 @Override
                 public void onFailure(Throwable caught) {
                     // This signals that the bundle does not yet exist
-                    if (caught instanceof BundleNotFoundException) {
-                        handleBundleNotFoundException((BundleNotFoundException) caught);
+                    BundleNotFoundException bnfe = unpackBundleNotFoundException(caught);
+                    if (bnfe != null) {
+                        handleBundleNotFoundException(bnfe);
 
                     } else {
                         // Escape it, since it contains the URL, which the user entered.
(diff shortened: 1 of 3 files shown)


rhqci:

pilhuhn:
@jshaughn Can you please look at this?

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/server/gwt/BundleGWTServiceImpl.java
```diff
@@ -71,7 +71,7 @@ public ResourceTypeBundleConfiguration getResourceTypeBundleConfiguration(int co
     public BundleVersion createOrStoreBundleVersionViaURL(String url, String username, String password)
         throws RuntimeException {
         try {
-            BundleVersion results = bundleManager.createBundleVersionViaURL(getSessionSubject(), url, username,
+            BundleVersion results = bundleManager.createOrStoreBundleVersionViaURL(getSessionSubject(), url, username,
```
josejulio: I would like to point out that I modified "createBundleVersionViaURL" (changed to  "createOrStoreBundleVersionViaURL") to match method invocation. Though I'm not sure if is the right thing to do or if we should leave it as "createBundleVersionViaURL".

jmazzitelli: I would NOT change API names like this. IIRC, that is a public API and if so I'm fairly certain customers will have used it. If anything, create a new method with the name you want, but maintain the old method for backward compatibility.

jmazzitelli: Unless I misunderstood your comment - when you say you "modified  createBundleVersionViaURL (changed to "createOrStoreBundleVersionViaURL") " I assume that meant you renamed the API. If you just changed what method you called and that is required, that's OK. I just thought you changed the public API name - in which case, not OK.

josejulio: Yeah, sorry if I didn't explained well. I'm just pointing to other method, the one that is *supposed* to be called instead.

When I say *supposed*, is because:
 - The method match the method invocation, and
 - The method *bundleManager.createOrStoreBundleVersionViaURL* had traces of the *token* logic that is used on the 'Upload'.

josejulio: Just to clarify: I'm not changing any public API.

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/bundle/create/BundleUploadDistroFileStep.java
```diff
@@ -297,6 +301,24 @@ public void onFailure(Throwable caught) {
             });
     }
 
+    private BundleNotFoundException unpackBundleNotFoundException(Throwable caught) {
+        if (caught instanceof RuntimeException) {
+            String message = caught.getMessage();
+            int patternStart = message.indexOf(BUNDLE_NOT_FOUND_EXCEPTION_PATTERN_START);
+            if (patternStart > 0) {
```
burmanm: Please use -1 to indicate something wasn't found, as now it could have been found from index 0

josejulio: Right, Will fix right away.

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/server/gwt/BundleGWTServiceImpl.java
jshaughn: Yes, this looks like the wrapper was then calling the wrong server-side method, I think the change is correct.

jshaughn:
Looks OK to me.

PR #196, opened 2015-12-16
Bug 1272358 - When creating a big bundle via UI the wizard shows erro…

josejulio:
…rs if user clicks Next button multiple times

Added a flag to know if the token processing(User doesn't have permission to add withou Bundle Group, so a token processing is needed) is and show a message, similar to the one when shows when click next while uploading a large bundle.


rhqci:

pilhuhn:
@jshaughn Can you have a look at this please?

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/bundle/create/BundleGroupsStep.java
```diff
@@ -62,11 +62,15 @@
     // a liberal timeout here...
     private static final int BUNDLE_CREATE_INITIAL_VERSION_TIMEOUT = 10 * 60 * 1000;
 
+    private static final int PROCESSING_IDLE = 0;
```
burmanm: Don't use ints to emulate enums

josejulio: Ok, will change that to actual enums.

jshaughn:
Looks OK to me

PR #197, opened 2015-12-17
Fixes incorrect versions for dependencies on OpenShift-plugin and rhq…

josejulio:
…-serverplugin-jdr.


rhqci:

PR #198, opened 2016-01-05
Bug 1257741 - First attempt of saving SNMP alert configurations is fa…

josejulio:
…iled if UDP transport protocol is used.

The property didn't had a default value, it was using "defaultValue" instead of "default".


rhqci:

PR #199, opened 2016-01-08
Bug 1184000 - Bundles missing from the left list / bundle navigation …

josejulio:
…tree

Removed the bundles limit.
Sorted the bundles by name.


rhqci:

PR #206, opened 2016-02-16
Adds support for wildlfy application server 10.

dmaehler:


rhqci:

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/JBossProductType.java
```diff
@@ -143,6 +144,8 @@ private static JBossProductType determineJBossProductTypeViaProductConfFile(File
                 productType = JBossProductType.BPMS;
             } else if (slot.equals("dv")) {
                 productType = JBossProductType.JDV;
+            } else if (slot.equals("wildfly-full")) {
```
pilhuhn: Is this correct? I think `-full` denotes the profile but not the server version. I think it may be required to also check the api version or such.
@burmanm wdyt?

dmaehler: This is the content that I found in bin/product.conf
```
# JBoss product branding configuration
slot=wildfly-full
```

I'm using WildFly 10.0.0.Final

burmanm: Hi, thanks for the patch. While you're right that they've added the "wildfly-full" as product.conf to Wildfly 10 (thus breaking a little bit our earlier detection code), this isn't quite enough. This will detect the Wildfly 10, but you'll receive a InvalidPluginConfiguration at some point since the validateServerAttributes() in BaseServerComponent will also fail:

```
[standalone@localhost:9990 /] read-attribute product-name
WildFly Full
[standalone@localhost:9990 /]
```

And this doesn't match "wildfly-full". Also, I did not check the other Wildfly versions if the web for example returns "wildfly-web" (and it if does, then adding support for Wildfly 10 should detect these as well correctly).

pilhuhn: Wildfly 7,8,9,10 (7=as7) can be detected by management-major-version iirc

wf 9 has management-major-version=3
wf 10 has management-major-version=4
as7 has 1 and I think wf8 had 2 (needs to be confirmed).

Also 

on wf 10
release-version=2.0.10.Final

wf9(.01)
release-version=1.0.1.Final

and so on, but I believe the management-major-version
is the most unique per major WF version.

dmaehler: I have had that exception before and the monitoring hasn't started, but now it's working fine on a RHQ 4.13, without any messages.

burmanm:
Thank you for the contribution. As we decided to split Wildfly 10 monitoring to a new plugin, I have added support for detecting Wildfly 10 installations there in commit 706a8dc5bb5ab5b4b7de73e8359c8460f4c428e7

PR #208, opened 2016-03-01
Bug 1308947 - Group Operation sequential execution list limited to 50…

josejulio:
… members.

Removed the limit of resources.


josejulio:
Updated, deferred the loading of member list till the user actually tries to change the execution order, thus the resource list won't be loaded on page load.

PR #209, opened 2016-03-04
Add resource definition or the build fails (Maven 3.3.9)

tsegismont:
If if I don't add this, the build fails because src/main/bundle resources don't get copied over to target/classes


tsegismont:
@burmanm would you mind to have a look?

PR #212, opened 2016-03-08
Bug 1311140 - EAP7 - missing rt filter modules for eap7

tsegismont:
Added new rtfilter-wfly-10-subsystem
Added new rtfilter-wfly-10-subsystem to downloads section
Fixed Web Runtime component (now at path subsystem=undertow)
Verified Calltime metric collection

### DIFF
--- a/modules/enterprise/server/appserver/src/main/downloadable-resources/connectors/README.txt
+++ b/modules/enterprise/server/appserver/src/main/downloadable-resources/connectors/README.txt
@@ -8,8 +8,8 @@ This is a servlet filter that can measure response times for HTTP requests
 made to servlets and write the response times to a logfile. The filter is
 compatible with any servlet 2.4 or later container running on Java 1.4 or 
 later. However, since upport for parsing the response time logfiles is
-only provided by the jboss-as, jboss-as-5, jboss-as-7, and tomcat RHQ 
-plugins, it typically is only of value to deploy the filter to JBoss AS
+only provided by the jboss-as, jboss-as-5, jboss-as-7, wfly-10, and tomcat RHQ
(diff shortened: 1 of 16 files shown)


tsegismont:
@burmanm would you mind to review ? Thanks

tsegismont:
@burmanm in particular, I'm interested in your opinion about the "Web Runtime" resource name. Should we keep it as is? Or change it to "Undertow Runtime"?

pilhuhn:
On 8 Mar 2016, at 16:10, Thomas Segismont wrote:

> @burmanm in particular, I'm interested in your opinion about the "Web 
> Runtime" resource name. Should we keep it as is? Or change it to 
> "Undertow Runtime"?

Undertow is imo an implementation detail.
"Web Runtime" is generic and not the name of the previous 
implementation.

burmanm:
Web Runtime was actually name of the previous implementation (JBossWeb runtime and indicating subsystem=web when now it is subsystem=undertow).

loleary:
@burmanm I think @pilhuhn is still correct.

Additionally, we should avoid using community project or implementation names for resource type names. This is due to the potential they will change. 

Undertow is a web server and therefore it would make sense that we would refer to it by its function/purpose.

burmanm:
The issue with that approach is that RHQ will still add the resource subsystem as "undertow" as it looks at the path (subsystem=undertow) and not the description saying "Web Runtime". So referring it as Web Runtime in some places, but keeping the configuration and metrics under "undertow" makes it a bit difficult in UI perspective.

That's the same reason why we don't have "messaging" subsystem with EAP7/Wildfly 10 anymore, but instead we have "messaging-activemq" there. It's just the way we're currently adding sub-resources.

PR #213, opened 2016-03-09
Bug 1313955 - JON agent does not recognize EAP7 Undertow subsystem that replaces EAP6 Web subsystem

josejulio:


modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -8709,245 +8709,251 @@
...
-
-    <service name="VHost"
-             discovery="SubsystemDiscovery"
-             class="VHostComponent"
-             createDeletePolicy="both">
+      
+      <service name="Buffer Cache"
+                            discovery="SubsystemDiscovery"
+                            class="BaseComponent"
+                            createDeletePolicy="both">
```
burmanm: Please check these, do we really allow creating these resources.

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -8709,245 +8709,251 @@
...
+                  </service><!-- End of host service -->
+      </service> <!-- End of Server service -->
+      <service name="Servlet Container"
+                  discovery="SubsystemDiscovery"
+                  class="BaseComponent"
+                  createDeletePolicy="both">
 
       <plugin-configuration>
-        <c:simple-property name="path" readOnly="true" default="virtual-server"/>
+          <c:simple-property name="path" readOnly="true" default="servlet-container=default"/>
```
burmanm: This is restricting the path to a single servlet-container. There can be multiple, so the "default" should not be enforced here.

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
josejulio: Could not find an option to add on from the EAP7 GUI, though there is a way for setting a servlet "Default buffer cache"

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
josejulio: Right, will change it.

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
burmanm: By default, do not allow adding / deleting in these subsystems. EAP7 GUI is not a good source for information, check what's returned by the CLI. If we later find out something is possible to be created, then add that ability.

josejulio: The CLI had the "add" and "delete" operation on that node, that's why i though it could be possible. Is that enough to decide if we should allow adding / deleting ? If not I'll remove the createDeletePolicy once we get more information.

burmanm: I'd say remove for now and lets revisit these later (those add operations might require some Java coding also)

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -8755,245 +8755,413 @@
...
+          
+          ./d2d.sh ... "subsystem=undertow,configuration=filter" connection-limit
+          yields
+           <c:simple-property name="max-concurrent-requests:expr" required="false" type="string" readOnly="false" description="Maximum number of concurrent requests"/>
+           <c:simple-property name="queue-size:expr" required="false" type="string" readOnly="false" description="Number of requests to queue before they start being rejected"/>
+        
+           Should I put a "configuration" service or is more like a group like...
+      <c:group name="child:configuration=filter" displayName="Connection limit">
+        <c:simple-property name="max-concurrent-requests:expr" required="false" type="string" readOnly="false" description="Maximum number of concurrent requests"/>
+        <c:simple-property name="queue-size:expr" required="false" type="string" readOnly="false" description="Number of requests to queue before they start being rejected"/>
```
josejulio: I'm not sure how to put this part, currently looking at the CLI. But I suppose it requires some coding.

PR #216, opened 2016-03-11
[BZ 1300727] Fix validateServerAttributes() for fallback case

burmanm:


tsegismont:
For some reason, the build failed in core-domain module...

tsegismont:
retest this please

modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/itest/AbstractServerComponentTest.java
```diff
@@ -107,8 +107,10 @@ public void testServerAttributeValidation() throws Exception {
             + "start() method due to invalid baseDir.");
 
         // Change the expectedRuntimeProductName property
+        boolean exceptException = true;
```
tsegismont: Typo?

tsegismont:
@burmanm could you sum up what is the intention? There is a lot comments in the related BZs... My understanding is that we want non identifiable EAP6-based products to be discovered as raw EAP6 servers, but it doesn't look like it is what this PR does. Thanks

modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/itest/AbstractServerComponentTest.java
burmanm: Yes, a typo. Shouldn't matter though too much.

burmanm:
This isn't the patch that's doing that, this is just fixing what was wrong in the original commit to the release/jon3.3.x (one extra line and incorrectly behaving validateServerAttributes()), see comment 8 on the ticket.

modules/plugins/jboss-as-7/src/test/java/org/rhq/modules/plugins/jbossas7/itest/AbstractServerComponentTest.java
tsegismont: Not too much indeed. But without the typo you can  understand at first glance :)

tsegismont:
Le 11/03/2016 21:53, Michael Burman a écrit :
> This isn't the patch that's doing that, this is just fixing what was
> wrong in the original commit to the release/jon3.3.x (one extra line and
> incorrectly behaving validateServerAttributes()), see comment 8 on the
> ticket.
>

Got it now. Thanks

tsegismont:
LGTM. Do you know what happened to the build? It failed twice while building core-domain... Maybe a PR has been merged before which wasn't quite right.

PR #218, opened 2016-03-15
Bug 1315466 - Agent is throwing Error while JON server executing Oper…

josejulio:
…ations on JBoss server.

Detects if error code is 503 and returns a failure operation result with "server not available" message.

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ASConnection.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ASConnection.java
@@ -417,9 +417,14 @@ public class ASConnection {
         try {
             operationResult = mapper.readTree(responseBody);
         } catch (IOException ioe) {
-            String failureDescription = "Failed to deserialize response to " + operation
-                + " to JsonNode - response status was " + statusAsString(statusLine) + ", and body was ["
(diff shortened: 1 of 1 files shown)


modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ASConnection.java
```diff
@@ -417,10 +417,16 @@ private JsonNode deserializeResponseBody(Operation operation, StatusLine statusL
...
             operationResult = mapper.readTree(responseBody);
         } catch (IOException ioe) {
-            String failureDescription = "Failed to deserialize response to " + operation
-                + " to JsonNode - response status was " + statusAsString(statusLine) + ", and body was ["
-                + responseBody + "]: " + ioe;
-            LOG.error(failureDescription);
+            String failureDescription = "";
+            if (statusLine.getStatusCode() == 503) {
+                failureDescription = "Server not available [" + statusAsString(statusLine) + "]";
+                LOG.info(failureDescription);
```
burmanm: It's still an error, operation failed

josejulio: Ok, will update shortly.

PR #219, opened 2016-03-15
Bug 1246083 - CLI operations don't work when Native Local Authentication is enabled and Native management API host is set to 0.0.0.0

tsegismont:
Removed nativeHost and nativePort settings from EAP7 servers config, as
the CLI now connects to the HTTP management port.

Updated discovery classes so that when '0.0.0.0' is discovered,
'127.0.0.1' is used
Updated component classes to log a warning if native/management host is set to
'0.0.0.0'
Changed defaults from 'localhost' to '127.0.0.1' in HostConfiguration
classes to avoid troubles when localhost is configured to something else
than loopback address

Note that, with these changes, new servers bound to '0.0.0.0' will no
longer have this address in the resource name, but '127.0.0.1' instead.
Resource keys are not impacted. Already inventoried EAP6 resources are left
untouched.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseProcessDiscovery.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseProcessDiscovery.java
@@ -213,10 +213,18 @@ public abstract class BaseProcessDiscovery implements ResourceDiscoveryComponent
         initLogEventSourcesConfigProp(logFile.getPath(), pluginConfig);
 
         HostPort managementHostPort = hostConfig.getManagementHostPort(commandLine, getMode());
+        if ("0.0.0.0".equals(managementHostPort.host)) {
+            LOG.debug("Discovered management host set to 0.0.0.0, falling back to 127.0.0.1");
+            managementHostPort.host = "127.0.0.1";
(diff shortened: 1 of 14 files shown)


tsegismont:
Please do not merge yet. I need to make sure this works fine in domain mode as well.

burmanm:
I think native host / native port were correctly working in domain mode, but not in standalone, so check that part..

tsegismont:
Le 15/03/2016 19:40, Michael Burman a écrit :
> I think native host / native port were correctly working in domain mode,
> but not in standalone, so check that part.

They didn't in standalone mode just because the native interface is no 
longer bound by default.
The CLI used to connect over the native interface but now it connects 
over HTTP with upgrade to remoting protocol.
As we introduced nativeHost and nativeHost just for CLI with localAuth 
mode, I believe we shouldn't keep these attributes any more. I verified 
that localAuth mode works on a standalone server with http upgrade, now 
I need to verify it does with a domain controller as well.

tsegismont:
@burmanm just verified, localAuth works fine as well with a domain controller and http ugrade to remoting

tsegismont:
Rebased on master

tsegismont:
Closing because we are going to separate this PR in two parts: one for EAP6 plugin, one for EAP7 plugin.

PR #220, opened 2016-03-15
Bug 1218129 - Calltime metrics sort does not work properly

rubenvp8510:
Removed formatted value from FIELD_TOTAL attribute, used the value in milliseconds and leave formatted total field just for UI presentation on FIELD_TOTAL_STRING.


rhqci:

tsegismont:
LGTM

tsegismont:
add to whitelist

PR #223, opened 2016-03-17
Bug 1295863 - The number of resources in All Groups/Compatible Groups…

josejulio:
… page is not correct all the time

We can't assume that the order is the same.


josejulio:
I'm still testing this one.

jshaughn:
This looks good to me.  I suppose we maybe could have maybe found a way to sort the explicit and implicit results, and kept the more efficient logic, but the sorting itself would add some complexity and some processing time.

PR #230, opened 2016-03-23
Bug 1320260 - Unknown attribute when trying to edit Transactions Subs…

josejulio:
…ystem of EAP7

Removed relative-to (replaced by object-store-relative-to) and path (replaced by object-store-path).


burmanm:
You should fix the comments to indicate the change also. object-store-path for example has a link to the removed properties: "By default the value is treated as relative to the path denoted by the 'relative-to' attribute." (jboss-cli should give you the new description)

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/TransactionsComponent.java
```diff
@@ -54,9 +57,31 @@ public void updateResourceConfiguration(ConfigurationUpdateReport report) {
...
+
+        boolean processIdUuid = Boolean.valueOf(config.getSimpleValue("process-id-uuid"));
+        if (processIdUuid) {
+            // Do not send the value of [process-id-socket-max-ports] as this should be 10 and could confuse the user
+            // when trying to change the value.
+            config.remove("process-id-socket-max-ports");
+        }
+
+        return config;
+    }
```
josejulio: When [process-id-uuid] is true, [process-id-socket-max-ports] should be undefined  (according to the attribute description), but the value by default is '10'. 

I'm not sure if we should *hide* in the UI the value of [process-id-socket-max-ports] or if this would generate more confusion.

PR #232, opened 2016-03-30
Bug 1320289 - Exception when trying to edit Module options from Security subsystem

rubenvp8510:
Because this subsystem  is deprecated, I just ignore the exception.

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ModuleOptionsComponent.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ModuleOptionsComponent.java
@@ -572,10 +572,18 @@ public class ModuleOptionsComponent extends BaseComponent implements Configurati
 
                 //locate specific node and populate the config
                 String loginModuleIdentifier = attribute + ":";
+
                 int index = path.indexOf(loginModuleIdentifier);
                 String loginModuleIndex = path.substring(index + loginModuleIdentifier.length());
                 String[] split = loginModuleIndex.split(",");
-                int lmi = Integer.valueOf(split[0]);//Ex 0,1,30
+
+                int lmi = -1;
+                try {
+                    lmi = Integer.valueOf(split[0]);//Ex 0,1,30
+                } catch (NumberFormatException e) {
+                    // We can't parse anything meaningful
+                    return;
+                }
                 Value valueObject = currentAttributeState.get(lmi);
 
                 //retrieve current config changes    


modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ModuleOptionsComponent.java
```diff
@@ -572,34 +572,39 @@ public void updateResourceConfiguration(ConfigurationUpdateReport report) {
...
+                    write.addAdditionalProperty("name", attribute);//attribute to execute on
 
-                //now complete the write operation by updating the value
-                write.addAdditionalProperty("value", (Object) currentAttributeState);
-                executeWriteAndGenerateAs7ServerUpdate(report, conf, write);
+                    //now complete the write operation by updating the value
+                    write.addAdditionalProperty("value", (Object) currentAttributeState);
+                    executeWriteAndGenerateAs7ServerUpdate(report, conf, write);
+                } catch(NumberFormatException e) {
+                    // Ignore it, the value just wasn't available
```
josejulio: Shouldn't it at least output something to the log?

burmanm: No, it should not (it already does, lets avoid that spam), as we do not have a process for these modules that are not supporting our indexing scheme. So ignoring them is fine.

However, the NumberFormatException catches too long section in my opinion, hiding potentially some other issues than just the index parsing. So the area of try should be reduced.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ModuleOptionsComponent.java
```diff
@@ -572,34 +572,45 @@ public void updateResourceConfiguration(ConfigurationUpdateReport report) {
...
-                //list current conf stated being defined.
-                String mapKey = (String) conf.getMap().keySet().toArray()[0];
-                PropertyMap mapType = (PropertyMap) conf.getMap().get(mapKey);
-                //insert update logic.
-                //set this new state passed in as state of the attribute.
-                LinkedHashMap<String, Object> currentOptions = new LinkedHashMap<String, Object>();
-                for (String propertyKey : mapType.getMap().keySet()) {
-                    currentOptions.put(propertyKey, ((PropertySimple) mapType.get(propertyKey)).getStringValue());
+                
+                Value valueObject=null;
```
burmanm: This type of formatting is not consistent with the rest of the code. Try to add some more whitespace like the previous code has, such as ``try {`` and ``} catch (..) {`` as well as ``valueObject != null``.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ModuleOptionsComponent.java
```diff
@@ -572,34 +572,45 @@ public void updateResourceConfiguration(ConfigurationUpdateReport report) {
...
+                    for (String propertyKey : mapType.getMap().keySet()) {
+                        currentOptions.put(propertyKey, ((PropertySimple) mapType.get(propertyKey)).getStringValue());
+                    }
 
-                //make these settings the new state for this part of the attribute.
-                valueObject.setOptions(currentOptions);
-                currentAttributeState.set(lmi, valueObject);
+                    //make these settings the new state for this part of the attribute.
+                    valueObject.setOptions(currentOptions);
+                    currentAttributeState.set(lmi, valueObject);
```
burmanm: What will happen here when lmi is -1 and valueObject is null? This scenario will happen if parsing fails. Is there a reason why we won't just do:

``
try {
  lmi = Integer.valueOf(split[0]);
} catch (NumberFormatException e) {
  // We can't parse anything meaningful
  return;
}

rubenvp8510: if the parse of the number fails both `lmi` and `valueObject` will be null and the code inside the `if (valueObject==null) ` won't be executed, but yes I think it's better to put the return, essentially because if I can't parse the number I can't do anything else on that method, so I changed it.

PR #233, opened 2016-03-31
Bug 1321941 - EAP7 - Undertow - new created Servlet Container not loaded

josejulio:
Creates JSP and Websockets settings nodes when trying to create Servlet Container.

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/UndertowComponent.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/UndertowComponent.java
@@ -1,0 +1,78 @@
+/*
+ * RHQ Management Platform
+ * Copyright 2012, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


josejulio:
I'm not sure if this is the way to go with this kind of scenario.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/UndertowComponent.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @author Josejulio Martínez
+ */
+public class UndertowComponent extends BaseComponent<ResourceComponent<?>> {
+
+    @Override
+    public CreateResourceReport createResource(CreateResourceReport report) {
+        String path = report.getPluginConfiguration().getSimpleValue("path", "");
+        if (path.equals("servlet-container")) {
+            report.setStatus(CreateResourceStatus.INVALID_CONFIGURATION);
```
burmanm: Why make it INVALID_CONFIGURATION and still proceed?

josejulio: I saw at the beginning of the CreateResourceDelegate[1], I though of it as a default value.

[1] https://github.com/rhq-project/rhq/blob/master/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/CreateResourceDelegate.java#L47-L51

PR #234, opened 2016-04-04
Bug 1313955 Undertow is not recognized in domain mode.

josejulio:


josejulio:
I'll wait for #233 to be merged (or this one) to update the XML.

PR #237, opened 2016-04-12
Bug 1314388 - Do not load max-pool-size when derive-size is set.

josejulio:
max-pool-size takes the default value, even when is undefined, making it harder for the user to update the configuration as it can't be set along with derive-size.

> 1) Set derive-size and unset max-pool-size
> 2) Save
> 3) Change something else (Timeout or Timeout Unit)
> 4) Save

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/Ejb3Component.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/Ejb3Component.java
@@ -43,10 +43,13 @@ public class Ejb3Component extends BaseComponent<BaseComponent<?>> {
 
         CompositeOperation cop = new CompositeOperation();
 
-        if((derive == null || derive.getStringValue() == null) && maxPoolSize != null) {
+        boolean deriveIsNull = (derive == null || derive.getStringValue() == null);
+        boolean maxPoolSizeIsNull = (maxPoolSize == null || maxPoolSize.getIntegerValue() == null);
+
+        if (deriveIsNull && !maxPoolSizeIsNull) {
             cop.addStep(new UndefineAttribute(address, DERIVE_ATTRIBUTE));
             cop.addStep(new WriteAttribute(address, MAX_POOL_SIZE, config.getSimpleValue(MAX_POOL_SIZE)));
-        } else if(derive != null && (maxPoolSize == null || maxPoolSize.getStringValue() == null)) {
+        } else if (!deriveIsNull && maxPoolSizeIsNull) {
             cop.addStep(new UndefineAttribute(address, MAX_POOL_SIZE));
             cop.addStep(new WriteAttribute(address, DERIVE_ATTRIBUTE, config.getSimpleValue(DERIVE_ATTRIBUTE)));
         } else {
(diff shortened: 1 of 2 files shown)


burmanm:
I'm torn between this (as I was while developing the original fix). On one hand, the max-pool-size is always returned by the EAP7 - so we mimick what the product does, but on the other hand the usability is reduced.

@loleary Do you have an opinion on this, which one from support's perspective is better way?

loleary:
I think in cases where one property depends on another property, there shouldn't be a default value set in the UI. Instead, any default should get applied by the configuration update code and/or the load configuration code when writing and reading the configuration respectively.

For example, max-pool-size should be displayed as `20` in the UI if `max-pool-size` is set to `undefined` or `null` and `derive-size` is set to `undefined` or `null` or `none`. The value is set when configuration is loaded, When configuration update is called, it will detect that `derive-size` is undefined/unset or set to `none` and that `max-pool-size` is `20` meaning that `max-pool-size` will get sent to the EAP server while `derive-size` will not.

If on the other hand, `derive-size` is set to a value other then `null` or _unset_ or `none` and `max-pool-size` is set to a value other then _unset_, perhaps the configuration update method can return a meaningful error letting the user know that when `derive-size` is set to a value other then `none`, `max-pool-size` must be _unset_.

I therefore think that the `defaultValue` attribute will need to be removed from `max-pool-size` and defaults should be determined based on existing configuration values and set by the appropriate component configuration load/update methods.

josejulio:
In this particular case, on EAP7, when you undefine `max-pool-size` and you query for its value afterwards, the value returned is `20`.
This arise a problem with the following use case:

1. Set `derive-size` and undefine `max-pool-size`
2. Save
3. Set `Timeout`
4. Save

Expected result: 
  Both `Save` operations are successful.
Actual result:
  First operation is success and the second fails because both `derive-size` and `max-pool-size` are set.

Other workaround for these kind of situations (that won't compromise the usability while mimicking what the product does) would be to check if `max-pool-size` value is the same, if so, remove it from the operation, thus it will only fail if the user explicitly changes the value of `max-pool-size`

loleary:
Isn't there some indication that this is the default value? Can we not use that?

In other words, if the value is `20` in the case of `max-pool-size`, we know it is the default. This might be risky though.

Otherwise, in your _workaround_ are you simply thinking of having config update load the configuration before applying it? To check to see if the current value are same/different then the configuration about to be applied?

josejulio:
I'm thinking that when `derive-size` is set I could verify the value of `max-pool-size`, if is the default one (need to check how to get this value) we could simply ignore it.

loleary:
Hmm, that seems like it would be confusing though. From a user perspective, I see `max-pool-size` as `20`, I would think my pool size is _20_. 

I think this is essentially the correct thing to do but `20` should not make it into the UI if `derive-size` is not `none` or `null` or _unset_. And if the user sets it to `20` and `derive-size` **is set**, I really think the safe thing is an error.

josejulio:
Great, that's what this PR does (preventing `max-pool-size` from making into the UI).

loleary:
But I think the plug-in descriptor still defines a default value of `20`. This will result in the `20` from always being displayed. Did you fix this?

josejulio:
Fixed that and now it detects when `max-pool-size` is not set.

PR #243, opened 2016-04-15
Bug 1327655 - Unable to configure welcome-file via JON

rubenvp8510:


modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -3227,7 +3227,8 @@
 
         <service name="Welcome File (Managed Server)"
                  discovery="SubsystemDiscovery"
-                 class="BaseComponent">
+                 class="BaseComponent"
+                 createDeletePolicy="both">
```
josejulio: Managed servers tend to be read-only.

PR #246, opened 2016-04-26
Bug 1288455 - The data aggregation job in JBoss ON stopped due to unr…

josejulio:
…eachable storage node

 - Moved permits acquisition inside AggregationTask.
 - Release permits upon exception on AggregationTask.
 - Added a timeout to DataAggregation.

### DIFF
--- a/modules/enterprise/server/server-metrics/src/main/java/org/rhq/server/metrics/aggregation/DataAggregator.java
+++ b/modules/enterprise/server/server-metrics/src/main/java/org/rhq/server/metrics/aggregation/DataAggregator.java
@@ -173,7 +173,7 @@ class DataAggregator<T extends NumericMetric> {
             }
             iterator = null;
             taskTracker.finishedSchedulingTasks();
-            taskTracker.waitForTasksToFinish();
+            taskTracker.waitForTasksToFinish(1, TimeUnit.DAYS);
         } catch (InterruptedException e) {
             log.warn("There was an interrupt while scheduling aggregation tasks.", e);
             taskTracker.abort("There was an interrupt while scheduling aggregation tasks.");
@@ -209,6 +209,7 @@ class DataAggregator<T extends NumericMetric> {
                         fetchData(batch, Bucket.SIX_HOUR);
                         processBatch(batch, Bucket.TWENTY_FOUR_HOUR);
                 }
+
             }
         });
         taskTracker.addTask();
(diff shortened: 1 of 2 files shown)


modules/enterprise/server/server-metrics/src/main/java/org/rhq/server/metrics/aggregation/DataAggregator.java
```diff
@@ -351,8 +351,10 @@ public AggregationTask(Batch batch) {
         @Override
         public void run() {
             try {
+                permits.acquire();
                 run(batch);
             } catch (Exception e) {
+                permits.release();
                 log.error("Aggregation will be aborted due to an unexpected error", e);
                 taskTracker.abort("Aborting aggregation due to an unexpected error: " + e.getMessage());
```
josejulio: I'm not sure if an exception here could be the culprit of that BZ, the logs never showed  "Aggregation will be aborted due...".

burmanm: Yea, it's not. I'm assuming it's still running the CQL command and that's just "stuck" or something along those lines.

modules/enterprise/server/server-metrics/src/main/java/org/rhq/server/metrics/aggregation/DataAggregator.java
```diff
@@ -351,8 +351,10 @@ public AggregationTask(Batch batch) {
         @Override
         public void run() {
             try {
+                permits.acquire();
                 run(batch);
             } catch (Exception e) {
+                permits.release();
```
burmanm: This can't be right? Only in the case of Exception would you release something here (and then in one FutureCallback finally blocks one more semaphore release). Thus a single acquire, two releases..

josejulio: The FutureCallback only gets scheduled at the very end[1] of the AggregationTask. 
If an exception is raised before reaching that it would release only. If the code that schedules the FutureCallback raises an exception after scheduling it would release twice. 

I could make sure to only release the semaphore if the operation doesn't get scheduled.

[1] https://github.com/josejulio/rhq/blob/bugs/1288455/modules/enterprise/server/server-metrics/src/main/java/org/rhq/server/metrics/aggregation/DataAggregator.java#L131

josejulio:
@burmanm Made some modifications on the code. WDTY? Should we remove those changes and only leave the timeout?

josejulio:
Finished testing and ready to be reviewed.

PR #247, opened 2016-04-29
Bug 1306602 - Uninventory of resource leaves orphaned content data in…

rubenvp8510:
… the database

### DIFF
--- a/modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
+++ b/modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
@@ -57,6 +57,9 @@ import org.rhq.core.domain.util.OSGiVersionComparator;
  */
 @Entity
 @NamedQueries({
+    @NamedQuery(name = PackageVersion.QUERY_FIND_ID_PACKAGES_BY_RESOURCE_ID, query = "" //
+        + "SELECT DISTINCT pv.generalPackage.id FROM PackageVersion pv WHERE pv.id IN "
+        + "(SELECT ip.packageVersion.id FROM InstalledPackage ip WHERE ip.resource.id = :resourceId )"),
     @NamedQuery(name = PackageVersion.QUERY_FIND_BY_PACKAGE_VERSION, query = "" //
         + "SELECT pv FROM PackageVersion AS pv " //
         + " WHERE pv.generalPackage.id = :packageId " //
@@ -139,7 +142,13 @@ import org.rhq.core.domain.util.OSGiVersionComparator;
         + "   AND pv.repoPackageVersions IS EMPTY " //
         + "   AND pv.installedPackages IS EMPTY " //
         + "   AND pv.installedPackageHistory IS EMPTY "),
-
+    @NamedQuery(name = PackageVersion.DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS, query = "DELETE PackageVersion pv "
+        + " WHERE pv.id NOT IN (SELECT pvcs.packageVersion.id " //
+        + "                       FROM PackageVersionContentSource pvcs) " //
+        + "   AND pv.repoPackageVersions IS EMPTY " //
+        + "   AND pv.installedPackages IS EMPTY " //
+        + "   AND pv.installedPackageHistory IS EMPTY " //
+        + "   AND pv.generalPackage.id = :packageId"),
     @NamedQuery(name = PackageVersion.DELETE_SINGLE_IF_NO_CONTENT_SOURCES_OR_REPOS, query = "DELETE PackageVersion pv "
         + " WHERE pv.id = :packageVersionId" //
         + "   AND pv.repoPackageVersions IS EMPTY " //
@@ -250,7 +259,7 @@ public class PackageVersion implements Serializable {
     // Constants  --------------------------------------------
 
     private static final long serialVersionUID = 1L;
-
+    public static final String QUERY_FIND_ID_PACKAGES_BY_RESOURCE_ID = "PackageVersion.findIdPackagesByResouceId";
     public static final String QUERY_FIND_BY_PACKAGE_VERSION = "PackageVersion.findByPackageVersion";
     public static final String QUERY_FIND_BY_PACKAGE_VER_ARCH = "PackageVersion.findByPackageVerArch";
 
@@ -268,6 +277,7 @@ public class PackageVersion implements Serializable {
     public static final String QUERY_FIND_METADATA_BY_RESOURCE_ID = "PackageVersion.findMetadataByResourceId";
     public static final String QUERY_GET_PKG_BITS_LENGTH_BY_PKG_DETAILS_AND_RES_ID = "PackageVersion.getPkgBitsLengthByPkgDetailsAndResId";
     public static final String DELETE_IF_NO_CONTENT_SOURCES_OR_REPOS = "PackageVersion.deleteIfNoContentSourcesOrRepos";
+    public static final String DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS = "PackageVersion.deleteByPkgIfNoContentSourcesOrRepos";
     public static final String DELETE_SINGLE_IF_NO_CONTENT_SOURCES_OR_REPOS = "PackageVersion.deleteSingleIfNoContentSourcesOrRepos";
     public static final String DELETE_MULTIPLE_IF_NO_CONTENT_SOURCES_OR_REPOS = "PackageVersion.deleteMultipleIfNoContentSourcesOrRepos";
     public static final String DELETE_PVPV_IF_NO_CONTENT_SOURCES_OR_REPOS = "PackageVersion.deletePVPVIfNoContentSourcesOrRepos";
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 
 import javax.ejb.EJB;
 import javax.ejb.Stateless;
@@ -58,6 +59,8 @@ import org.quartz.Scheduler;
 import org.quartz.SchedulerException;
 import org.quartz.SimpleTrigger;
 
+import org.jboss.ejb3.annotation.TransactionTimeout;
+
 import org.rhq.core.db.DatabaseType;
 import org.rhq.core.db.DatabaseTypeFactory;
 import org.rhq.core.domain.alert.Alert;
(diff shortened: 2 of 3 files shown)


modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
```diff
@@ -139,7 +142,13 @@
...
         + "   AND pv.installedPackages IS EMPTY " //
         + "   AND pv.installedPackageHistory IS EMPTY "),
-
+    @NamedQuery(name = PackageVersion.DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS, query = "DELETE PackageVersion pv "
+        + " WHERE pv.id NOT IN (SELECT pvcs.packageVersion.id " //
+        + "                       FROM PackageVersionContentSource pvcs) " //
+        + "   AND pv.repoPackageVersions IS EMPTY " //
+        + "   AND pv.installedPackages IS EMPTY " //
+        + "   AND pv.installedPackageHistory IS EMPTY " //
+        + "   AND pv.generalPackage.id IN ( :packageIds )"),
```
jshaughn: My memory may be failing me but I'm not totally sure this is safe because BundleFile is, I think, also backed by PackageVersion.   Still, it may be OK because bundle packages and resource packages should likely be independent.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -783,6 +789,62 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
...
+
+        String[] namedQueriesToExecute = new String[] { //
+        PackageVersion.DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS, //
+            PackageBits.DELETE_IF_NO_PACKAGE_VERSION };
+
+        for (String namedQueryToExecute : namedQueriesToExecute) {
+            try {
+                Query namedQuery = entityManager.createNamedQuery(namedQueryToExecute);
+                if (namedQueryToExecute.equals(PackageVersion.DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS)) {
+                    namedQuery.setParameter("packageIds", packages);
```
jshaughn: I think this will fail on Oracle if there are more than 1000 packages.  Oracle has an IN clause limit of 1000. The same could be true of getInstalledPackages(resourceIds) but actually it is OK because this logic is invoked on a single resource.  So, the limit is not an issue for getInstalledPackages and I doubt a single resource has that many packages.

loleary: Actually, as Jay indicated, this will fail on Oracle due to the 1000 element limit for IN criteria. It is also not impossible for a resource to have 1000 or more packages. There could be a 100 WARs deployed to an EAP instances with 15 versions each. This would equate to 1500 packages.

burmanm:
I don't like this approach of using these hand-made batches in the client side.

Assuming we're inside a transaction (in Oracle/Postgres), it does not matter if we use a "1000 limit IN" or just do all the deletes as single DELETE statements.

The performance issues come from the fact that we have to do open transaction, statement, close transaction. By doing 2000 single DELETE statements in a single transaction is faster than doing 2 times 1000 DELETE IN. The operation and log writes do not differ even if we use a single command to delete all the lines with a single SQL command or with multiple, as long as it's in the same transaction.

Also, the "IN" clause is silly. Why aren't we just using a single JOIN command here to do all the selection on the DB side, instead of fetching them to the client, mapping memory and then sending them back to the DB? Avoids all the issues.

rubenvp8510:
I don't like either, but I need to take this approach because the only way to get a relation between `Resource` and the `Package` is using the `rhq_installed_package` table.

In the deletion logic I can't delete the `Packages` and `PackageVersions` before delete the `InstalledPackages`, this is because of the database Foreign Key constraints, so I need to do a query first to get the packages associated with the resource, and then start the deletion (validating if that package doesn't have a repository associated too), is in this case that I need to retain it in memory.

If I delete the `InstalledPackages` first I'll lose the only relationship that we have between packages and resources.

Also I don't think we'll have a lot of batches here, so I think putting it on a transaction could solve the performance problem

@burmanm  WDYT?

burmanm:
Single transaction is fine for a very large amount of data, so that's not an issue. So just ditch the "1000" per batch logic as it's messy and do it one by one (still in the single transaction).

loleary:
Well, from a single transaction aspect, keep in mind that the transaction
timeout is normally 10 minutes or less. This means that if the delete will
take longer then that, it will always fail.

Would it make more sense to just use a native query here?

metlos:
Note that batching by 1000 is necessary if you're doing IN (... ids ...), because Oracle can't handle more, AFAIK.

burmanm:
Yes, but you can avoid the 'IN' by using single statements (the performance is the same). And in this case the transaction boundary is the same (so doesn't solve Larry's issue, as the transaction starts at the method)

jshaughn:
I think this PR identifies and addresses the causes of the issue.  But I think we may be able to work in a fix that is more in line with the existing code.  I agree with @burmanm that there is no need for extra transactioning, although if you look closely you'll see we actually perform a large # of new transactions already, mainly to avoid locking issues, I think.  The newly introduced cleanPackageBits and cleanPackageVersions both declare new txs, and that may not be necessary.  But more than that, I'm not sure that we need to perform this work outside of the bulk query approach we already employ.   We are still only performing sync uninventory one resource at a time, but @loleary is right that we should beware of the default timeout.  I doubt we'd hit it for one resource but we can increase it for this method, which runs async anyway.  I think we can probably do this more like the way we do the bulk deletes of the other entities hanging off of Resource. The queries are a little bit verbose, but not too bad, I think.  Take a look at this branch and feel free to comment, I have not tested this but I think it should work...

https://github.com/rhq-project/rhq/compare/master...jshaughn:bz1306602?expand=1

jshaughn:
OK, after talking with Ruben it's clear that my attempt to work this into the existing structure is not going to work.  The FK relationships aren't going to allow it to be done.  I think his approach is the right one.  I'll go back now and just make some comments on this original PR...

modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
```diff
@@ -139,7 +142,13 @@
         + "   AND pv.repoPackageVersions IS EMPTY " //
         + "   AND pv.installedPackages IS EMPTY " //
         + "   AND pv.installedPackageHistory IS EMPTY "),
-
+    @NamedQuery(name = PackageVersion.DELETE_BY_PKG_IF_NO_CONTENT_SOURCES_OR_REPOS, query = "DELETE PackageVersion pv "
+        + " WHERE pv.id NOT IN (SELECT pvcs.packageVersion.id " //
+        + "                       FROM PackageVersionContentSource pvcs) " //
```
jshaughn: Perhaps IN clause could be avoided by replacing with:
  NOT EXISTS ( SELECT * FROM PackageVersionContentSource pvcs WHERE pvcs.packageVersion.id = pv.id )

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -764,6 +766,10 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
 
         boolean hasErrors = false;
         boolean debugEnabled = LOG.isDebugEnabled();
+
+        // Get all packages installed in this resource
```
jshaughn: I'd like to see more detail in this comment to explain why this is different than everything else:

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -764,6 +766,10 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
...
+        List<Integer> installedPackages = getInstalledPackages(resourceId);
+
```
jshaughn: Because this is a list of packageIds I think it would be more clear if the variable name were 'installedPackageIds'

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -783,6 +789,82 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
...
+        // If this resource had packages installed, remove their version and bits
+        if (installedPackages.size() > 0) {
+            hasErrors |= cleanOrphanedPackages(overlord, installedPackages);
+        }
+
+        return hasErrors;
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<Integer> getInstalledPackages(Integer resourceId) {
```
jshaughn: perhaps getInstalledPackageIds()

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -783,6 +789,82 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
...
+
+        List<Integer> installed_packages = new ArrayList<Integer>();
```
jshaughn: better to stick with camelCase, I think.   And again, installedPackageIds would be more clear.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerBean.java
```diff
@@ -783,6 +789,82 @@ private boolean uninventoryResourceBulkDeleteAsyncWork(Subject overlord, int res
                 resourceIds);
         }
 
+        // If this resource had packages installed, remove their version and bits
+        if (installedPackages.size() > 0) {
+            hasErrors |= cleanOrphanedPackages(overlord, installedPackages);
```
jshaughn: If we are not going to remove the Package (if it has no more PackageVersions) then this should probably be renamed cleanOrphanedPackageVersions().

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/ResourceManagerLocal.java
```diff
@@ -499,4 +499,26 @@ Resource getResourceByParentAndKey(Subject user, @Nullable
...
+
+    /**
+     * INTERNAL
+     *
+     * Remove all orphaned package bits
+     *
+     * @return
+     */
+
+    boolean cleanPackageBits();
```
jshaughn: This jdoc is really just clutter, I'd remove it and the other one as well.

jshaughn:
Last comment, to address the concern of a Tx timeout, You could add:

rubenvp8510:
I fixed some code style issues and put the TransactionTimeout annotation, This is ready for another review.

I can squash the commits once is ok for merge.

PR #248, opened 2016-05-03
Bug 1331538 - Error when trying to save RemoteConnectionFactory confi…

josejulio:
…guration


josejulio:
Needs some changes before merging.

josejulio:
Replaced with #274

PR #251, opened 2016-05-06
Bug 1333929 - Error when trying to save "Connection Factory" configur…

josejulio:
…ation on Profile.

  Also checks if the (Profile) sufix exists.
  Skips some checks on JMS Queue and JMS Topic
  Changed to readonly two attributes of JMS Queue.


modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/BaseComponent.java
```diff
@@ -954,4 +956,13 @@ public static ComplexRequest create(String requestName) {
     public static Log getLog() {
         return LOG;
     }
+
+    public static String resourceTypeNameByRemovingProfileSufix(String resourceTypeName) {
```
burmanm: Suffix, not Sufix ;)

josejulio: Yep

PR #253, opened 2016-05-13
Bug 1333732 - The EAP 7 plug-in needs to be able to identify JDG 7

josejulio:


burmanm:
Don't merge, this isn't enough to do the detection correctly.

PR #254, opened 2016-05-14
Change JBossProductType detection to apiVersion only

burmanm:
Remove Wildfly8 support, change AS7 plugin to only accept apiVersion 1.x and Wildfly10 plugin to accept only apiVersion 4.x. Also add JDG7 detection to wfly-10 plugin.


burmanm:
(Don't merge yet)

spinder:
@burmanm, On second look this change to EAP 6/jboss-as-7 appears more extensive than I originally thought.  This will mean a lot more extensive retesting by JON and all layered products and should go into EAP plugin pack update 05 at a minimum, and not just a one-off plugin pack update for JDG 7.  Could we instead require all products based upon EAP7, to change/use a different product type as EAP 7 did?  Seems more organized and we might be able to make it a requirement of adding EAP 7 plugin support for layered products. Would that be enough?

PR #256, opened 2016-05-23
Bug 1302800 - Remote agent auto upgrade failed when sslsocket encrypt…

josejulio:
…ion enabled

  Added new server service to send the public agent update endpoint address.

### DIFF
--- a/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java
+++ b/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java
@@ -1551,6 +1551,24 @@ public class AgentMain {
 
                                 FailoverListComposite failover_list = null;
                                 try {
+                                    // Verifies if the transport method is a (ssl)socket
+                                    // If so, requests the public agent update endpoint (in case we need to update to register)
+                                    try {
+                                        if (m_configuration.getServerTransport().contains("socket")) {
+                                            String newPublicAgentUpdateEndpoint = remote_pojo.getPublicAgentUpdateEndpointAddress();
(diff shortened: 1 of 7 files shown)


modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java
```diff
@@ -1551,6 +1551,24 @@ public void run() {
...
+                                            String oldPublicAgentUpdateEndpoint = m_configuration.getPublicAgentUpdateEndpointAddress();
+                                            if (oldPublicAgentUpdateEndpoint == null || !oldPublicAgentUpdateEndpoint.equals(newPublicAgentUpdateEndpoint)) {
+                                                m_configuration.setPublicAgentUpdateEndpointAddress(newPublicAgentUpdateEndpoint);
+                                            }
+                                        }
+                                    } catch (Exception exception) {
+                                        if (LOG.isDebugEnabled()) {
+                                            LOG.debug(exception, "Unable to get the PublicAgentUpdateEndpointAddress from server.");
+                                        } else {
+                                            LOG.info("Unable to get the PublicAgentUpdateEndpointAddress from server.");
```
burmanm: WARN/ERROR rather? As this prevents some functionality from working.

loleary: Agreed. I think the log message in the else block should be WARN.
--
Larry

josejulio: Right, I'll change to WARN. Thanks for checking.

PR #258, opened 2016-05-26
Fixing the NPE in the UI when adding or creating a user.

Jiri-Kremser:
The `uiCustomizationGrid` was null when creating new user or editing a user that wasn't logged in. This is because the show/hide subsystem configuration is available only to those users that are logged in.

That led to NPE when calling `uiCustomizationGrid.getRecords()` in the post save action that called `getIntegerList()`.


Jiri-Kremser:
@burmanm Could you please review and merge? It's the bug we were talking about on IRC.

PR #262, opened 2016-05-31
Bug 1212495 - Solaris10-Error in server log after Generate JDR Report…

josejulio:
… operation

Removes the fork to avoid low swap memory conditions.

That seems to be a problem with not enough *swap* space [1] [2]. 
A solution is to remove that chmod call, while still having the permissions set to 600.

[1] https://docs.oracle.com/cd/E19455-01/806-1075/msgs-1831/index.html
[2] https://community.oracle.com/thread/2458428?tstart=0


josejulio:
I still need to test on Solaris 10, though the exception should not happen anymore as I removed the chmod call.

josejulio:
This PR is ready for review.

Tested on Solaris 10. The exception is gone and the file permissions are as they should be.

PR #265, opened 2016-06-03
Bug 1213812 - After upgrade from JBoss ON 3.2 to 3.3, some rhq column…

rubenvp8510:
… families are unavailable and compaction operations fail

I marked the obsolete column families `one_hour_metrics`, `six_hour_metrics`, `twenty_four_hour metrics` and `metrics_index` as uninventoried.

There is no need to create a new operation, this is because when the discovery process adds the new column families the operation  automatically applies to the new column families resources discovered under the group.

I tested it and it seems to work well, but I'm not sure about that is the right place to put this code, I run this process as part of storage schema upgrade

` ./rhqctl upgrade --storage-schema `


jsanda:
LGTM

PR #267, opened 2016-06-09
Bug 1328316 - Required fields in map-property does not prevent finish…

rubenvp8510:
…ing the Resource Create Wizard


modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/factory/ResourceFactoryConfigurationStep.java
```diff
@@ -109,6 +109,7 @@ public void onFailure(Throwable throwable) {
 
                         private void createAndAddConfigurationEditor(ConfigurationDefinition def) {
                             editor = new ConfigurationEditor(def, startingConfig);
+                            editor.enableMapValidations(true);
```
burmanm: Could you describe situations where we don't want the maps to be validated?

rubenvp8510: The thing is, that component is used in a lot of parts of the UI, I just did it  as a precaution, trying to minimize the impact on the other parts of the code, but you're right probably I should remove it. I tested as far as I could and it seems that all works well.

PR #270, opened 2016-07-18
[BZ 1213812] Add clear columnfamilies option to rhqctl upgrade

burmanm:
If no storage-schema update is needed, use this one to run only the removeObsoleteColumnFamilies..


rubenvp8510:
LGTM

PR #273, opened 2016-10-24
Bug 1379834 Prevent override plugin config user values on resource version upgrade

rubenvp8510:
This PR address the BZ 1379834, it patches the InventoryManager on the agent side.

Overrides if and only if is a read-only property, this prevents to override user values when there are no plugin default values or the values are dynamic 

I tested this. and It works for me, this is ready for review.

### DIFF
--- a/modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
+++ b/modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
@@ -61,6 +61,7 @@ import org.rhq.core.clientapi.agent.metadata.PluginMetadataManager;
 import org.rhq.core.clientapi.agent.metadata.ResourceTypeNotEnabledException;
 import org.rhq.core.clientapi.agent.upgrade.ResourceUpgradeRequest;
 import org.rhq.core.clientapi.agent.upgrade.ResourceUpgradeResponse;
+import org.rhq.core.clientapi.descriptor.configuration.SimpleProperty;
 import org.rhq.core.clientapi.server.configuration.ConfigurationServerService;
 import org.rhq.core.clientapi.server.discovery.DiscoveryServerService;
 import org.rhq.core.clientapi.server.discovery.InvalidInventoryReportException;
@@ -69,7 +70,9 @@ import org.rhq.core.clientapi.server.discovery.StaleTypeException;
 import org.rhq.core.domain.configuration.Configuration;
 import org.rhq.core.domain.configuration.ConfigurationUtility;
 import org.rhq.core.domain.configuration.Property;
+import org.rhq.core.domain.configuration.PropertySimple;
 import org.rhq.core.domain.configuration.definition.ConfigurationDefinition;
+import org.rhq.core.domain.configuration.definition.PropertyDefinition;
 import org.rhq.core.domain.discovery.AvailabilityReport;
 import org.rhq.core.domain.discovery.MergeInventoryReportResults;
 import org.rhq.core.domain.discovery.MergeInventoryReportResults.ResourceTypeFlyweight;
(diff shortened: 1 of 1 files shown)


modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
```diff
@@ -1928,24 +1931,24 @@ private Configuration mergeExistingResourcePluginConfiguration(Resource resource
...
         boolean configChanged = false;
 
         // for each property, update the existing plugin config if discovery has set a non-default value
         for (String propertyName : pluginConfig.getAllProperties().keySet()) {
             Property discoveredProp = pluginConfig.get(propertyName);
-            Property defaultProp = defaultPluginConfig.get(propertyName);
-            if (!discoveredProp.equals(defaultProp)) {
+            // Only overrides the property if it is a read-only property
+            // See BZ: 1379834
+            if (propDefs.get(propertyName).isReadOnly()) {
```
josejulio: This looks well but I'm not sure how this will affect other plugins.

Maybe we could define an attribute on the XML. Something like `override_strategy` that let you choose how to globally override the attribute values, DEFAULT, READ_ONLY, ONLY_NOT_NULL_DEFAULT, etc.

Or maybe i'm just being paranoid and no one cares/uses this.

rubenvp8510: That sounds good, but I'm not sure because it adds more complexity on this logic, and it's only invoked in cases when there is an upgrade version.

loleary: Although the _override strategy_ idea is a good one, it isn't something that we should really be doing in the maintenance stream. Not only would it force a version bump in the plug-in XSD, it would require this new tag/attribute to be defined for many resources and their configuraitons.

I do not think there should be any harm in only applying the the change to read-only properties. We are already doing that. All we are saying now is that we do not want to replace user provided configuration.

However, are we sure that we also do not want to verify the value requires updating? For example:
```
-            if (!discoveredProp.equals(defaultProp)) {
+            // Only overrides the property if it is a read-only property
+            // See BZ: 1379834
+            if (propDefs.get(propertyName).isReadOnly() && !discoveredProp.equals(defaultProp)) {
```
Of course, I am still not sure what the purpose of `discoveredProp.equals(defaultProp)` is even for, considering that discovery should always discover a different property then the default.

rubenvp8510: the purpose of `!discoveredProp.equals(defaultProp)` Is supposed to check if the newly discovered property is different from the default, then, if it is different updates with the new value, otherwise, it doesn't update. But in most of the cases I saw, the default property is null, (that is because the default values are taken from defaultValue XML attribute) so the condition is always true. But we can still check for that , just in case.

josejulio:
@burmanm could you please review this? We think this approach is OK to solve the issue.

modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
josejulio: IIRC `defaultValue` is deprecated and fails if used on JON. Should return null on all the cases, right?

rubenvp8510: Yes, the problem is that `defaultProp` takes the value from  `defaultValue` attribute on XML plugin descriptor, and that attribute is not used anymore,  that is why I think it doesn't make sense to validate it. but we can leave the validation if we feel that is safer.

josejulio:
I thinks this one is ready to be merged unless any concerns from @loleary or  @burmanm.

PR #276, opened 2016-11-30
[BZ 1381720] Clean orphaned large objects

burmanm:
Add a new operation to clean orphaned large objects from the database

### DIFF
--- a/modules/plugins/postgres/src/main/java/org/rhq/plugins/postgres/PostgresDatabaseComponent.java
+++ b/modules/plugins/postgres/src/main/java/org/rhq/plugins/postgres/PostgresDatabaseComponent.java
@@ -19,18 +19,14 @@
 
 package org.rhq.plugins.postgres;
 
-import static org.rhq.core.domain.measurement.AvailabilityType.DOWN;
-import static org.rhq.core.domain.measurement.AvailabilityType.UP;
-import static org.rhq.core.domain.resource.CreateResourceStatus.FAILURE;
-import static org.rhq.core.domain.resource.CreateResourceStatus.SUCCESS;
-import static org.rhq.plugins.postgres.PostgresDiscoveryComponent.buildConnection;
-
 import java.sql.Connection;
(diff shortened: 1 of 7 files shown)


loleary:
This new operation will only be reliable on PostgreSQL 8.4 and earlier. I suggest we make a slight modification to make it compatible with all supported versions. Otherwise, the operation description should make it clear that it will only work if you have super-user role or if on version 8.4 or earlier.

modules/plugins/postgres/src/main/java/org/rhq/plugins/postgres/PostgresDatabaseComponent.java
```diff
@@ -342,4 +346,80 @@ private OperationResult invokeSql(Configuration parameters) throws SQLException
...
+     * @return OperationResult
+     */
+    private OperationResult vacuumLo() {
+        Connection c = null;
+        Statement clearVacuumL = null;
+        Statement unlinkStatement = null;
+        ResultSet rs = null;
+        try {
+            c = getPooledConnectionProvider().getPooledConnection();
+            PreparedStatement tempTable = c.prepareStatement("CREATE TEMP TABLE vacuum_l AS (SELECT DISTINCT loid AS lo FROM pg_largeobject)");
```
burmanm: Well, it works with 9.5.5 at least, if the user owns the database in question (I have rhqadmin user that is the owner of rhqdev database).

Do we have a use-case where the rhq-user has access to modify the schema, but not to access pg_largeobject ?

loleary: All standard use-cases prevent access to pg_largeobject by a non-superuser. This is true as of PostgreSQL 9.0 and later.

`ERROR: permission denied for relation pg_largeobject`

I am assuming in your testing you have given admin/superuser rights to your rhqadmin database user or have disabled permission checking altogether or have given explicit grants to LARGE OBJECT?

PR #278, opened 2016-12-13
Bug 1369101 - Only initialize securityServices if using sslsocket.

josejulio:


josejulio:
@jmazzitelli do you recall any side effect of only calling [initializeSecurityServices](https://github.com/josejulio/rhq/blob/6f3df21026d7dec585b7459a10f808afc325d0a2/modules/enterprise/comm/src/main/java/org/rhq/enterprise/communications/ServiceContainer.java#L1355-L1401) when using sslsocket ? 

As far as I could see that only inits sslsocket stuff.

josejulio:
@burmanm could you please review?

jmazzitelli:
On the face of it, this looks OK. The only thing I would check is to see if that MBean always needs to exist, no matter what transport protocol is used. See line 1397 where it registers the MBean. It probably is OK, but just something to keep in mind. Just make sure that MBean isn't required to always exist.

burmanm:
If the MBean was always required, then it would fail when running server in unsecured mode (this is only ran when security is enabled - even before the patch)

PR #279, opened 2016-12-13
Bug 1367505 - Trims result to match page controls.

josejulio:
There was a custom *search* that works fetching all the data.
It was missing the pagination of the results.


josejulio:
@rubenvp8510 could you please review?

PR #280, opened 2016-12-13
Bug 1399125 - Don't load every version configuration, just the one needed.

josejulio:


josejulio:
@burmanm could you please review?

PR #281, opened 2016-12-15
Bug 1382649 -Attempt to add notification for the second alert..

rubenvp8510:
Bug 1382649  Attempt to add notification for the second alert definition in JBoss ON UI throws message "Failed to get notification configuration preview".

  
Removed recursive AlertDefinition seems to solve the problem of deserialization for GWT.


josejulio:
I'm not sure why, but it does indeed seem to solve the serialization problem.
Lets ask @burmanm if he knows why when he comes back.

burmanm:
I guess it's the GWT bug 8481

burmanm:
As the workaround works, this is fine (updating GWT is much more difficult)

PR #283, opened 2017-01-04
Bug 1382041 - Try to attach JVM, even if credname is not present.

rubenvp8510:
I wasn't manage to reproduce the process without credname, but in this PR I tried to attach the JVM even if the credname is not present

I'm not sure if this is correct? WDYT @josejulio  @burmanm


burmanm:
.. which BZ?

josejulio:
[BZ-1382041](https://bugzilla.redhat.com/show_bug.cgi?id=1382041)

burmanm:
The attach API will not work if the user running the target PID is not the same as the agent. You can't access (with attach API) the other user's processes, as the unix socket prevents this. 

You can fix the NPE by making sure that vmUserName != null && equals (or actually, just changing the equals order to avoid asking null.equals)

burmanm:
The better way would of course be to check the uid of both processes instead of the userName.

burmanm:
Fixed in the master by commit 87208f971b4e78c6236abcc9ea50c4ec34825825

PR #284, opened 2017-01-10
WARN when setting rhq.server.tomcat.security.client-auth-mode to a va…

josejulio:
…lue different than true or false.

### DIFF
--- a/modules/enterprise/comm/src/main/java/org/rhq/enterprise/communications/ServiceContainer.java
+++ b/modules/enterprise/comm/src/main/java/org/rhq/enterprise/communications/ServiceContainer.java
@@ -1246,6 +1246,15 @@ public class ServiceContainer {
             if (transport.equals("https")) {
                 connector_config.put("SSLImplementation", RemotingSSLImplementation.class.getName());
             }
+
+            if (transport.equals("sslservlet") || transport.equals("https")) {
+                String tomcatAuthMode = "rhq.server.tomcat.security.client-auth-mode";
+                String tomcatAuthModeValue = System.getProperty("rhq.server.tomcat.security.client-auth-mode");
+                if (!tomcatAuthModeValue.equals("true") && !tomcatAuthModeValue.equals("false")) {
+                    LOG.warn(CommI18NResourceKeys.SERVICE_CONTAINER_CONFIGURATION_INVALID_TOMCAT_CLIENT_AUTH,
+                            tomcatAuthMode, tomcatAuthModeValue, "true", "false");
+                }
+            }
         }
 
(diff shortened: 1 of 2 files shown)


josejulio:
@rubenvp8510 could you review?

josejulio:
or actually, @burmanm could you review? Rubén doesn't have merge permissions.

modules/enterprise/comm/src/main/java/org/rhq/enterprise/communications/ServiceContainer.java
```diff
@@ -1246,6 +1246,15 @@ private void setupServerConnector() throws Exception {
             if (transport.equals("https")) {
                 connector_config.put("SSLImplementation", RemotingSSLImplementation.class.getName());
             }
+
+            if (transport.equals("sslservlet") || transport.equals("https")) {
+                String tomcatAuthMode = "rhq.server.tomcat.security.client-auth-mode";
+                String tomcatAuthModeValue = System.getProperty("rhq.server.tomcat.security.client-auth-mode");
+                if (!tomcatAuthModeValue.equals("true") && !tomcatAuthModeValue.equals("false")) {
```
burmanm: Are we on purpose case sensitive?

josejulio: This value is passed directly to jboss-as.
Couldn't find much info but it seems that these values are in lower case.
(From the docs): `When using sslservlet set rhq.server.tomcat.security.client-auth-mode parameter to true`
[BZ-1369081](https://bugzilla.redhat.com/show_bug.cgi?id=1369081#c5)

and found [this](https://github.com/wildfly/wildfly/blob/6b61a6003f704221f66dcd9f418bcb7af88fb9ab/legacy/web/src/main/java/org/jboss/as/web/WebMigrateOperation.java#L1042-L1065)

josejulio: So, yes, is on purpose.

burmanm: Should we mention that in the error message (it's not necessarily clear to anyone why ``TRUE`` is wrong and we propose ``true`` ? Otherwise looks good.

josejulio: I updated it to "The [{0}] preference specified is invalid [{1}] - it must be one of [{2}, {3}] (case sensitive)."

PR #285, opened 2017-01-13
Bug 1368107 - Increase the amount of items per page.

josejulio:
This increases the number of items per page on every gui/portal-war
that uses the PageControlSettingsUIBean.

### DIFF
--- a/modules/enterprise/gui/portal-war/src/main/java/org/rhq/enterprise/gui/common/paging/PageControlSettingsUIBean.java
+++ b/modules/enterprise/gui/portal-war/src/main/java/org/rhq/enterprise/gui/common/paging/PageControlSettingsUIBean.java
@@ -24,8 +24,30 @@ import javax.faces.model.SelectItem;
  * @author Joseph Marques
  */
 public class PageControlSettingsUIBean {
-    private SelectItem[] pageSizes = new SelectItem[] { new SelectItem("15", "15"), new SelectItem("30", "30"),
-        new SelectItem("45", "45") };
+    private SelectItem[] pageSizes;
+    private static String MAX_ITEMS_PER_PAGE = "rhq.server.gui.max-items-per-page";
+
+    public PageControlSettingsUIBean() {
+        try {
+            String maxItemsPerPage = System.getProperty(MAX_ITEMS_PER_PAGE);
+            int maxItemsPerPageInt;
+            if (maxItemsPerPage != null && (maxItemsPerPageInt = Integer.parseInt(maxItemsPerPage)) >= 45) {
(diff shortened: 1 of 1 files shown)


josejulio:
Not sure if this is the proper way to do it.

The user needs to add `rhq.server.gui.max-items-per-page` property to `rhq-server.properties` file.
Also this will propagate to every UI that uses `PageControlSettingsUIBean`.
If `rhq.server.gui.max-items-per-page` is not set, it will use the default values.

Do you consider that I should only modify the requested view ? (Content -> New)
(it seems) that can be done as easily as this.
@spinder @burmanm

modules/enterprise/gui/portal-war/src/main/java/org/rhq/enterprise/gui/common/paging/PageControlSettingsUIBean.java
```diff
@@ -24,8 +24,25 @@
...
-    private SelectItem[] pageSizes = new SelectItem[] { new SelectItem("15", "15"), new SelectItem("30", "30"),
-        new SelectItem("45", "45") };
+    private SelectItem[] pageSizes;
+    private static String MAX_ITEMS_PER_PAGE = "rhq.server.gui.max-items-per-page";
+
+    public PageControlSettingsUIBean() {
+        try {
+            String maxItemsPerPage = System.getProperty(MAX_ITEMS_PER_PAGE);
+            if (maxItemsPerPage != null && Integer.parseInt(maxItemsPerPage) > 45) {
+                pageSizes = new SelectItem[] { new SelectItem("15", "15"), new SelectItem("30", "30"),
```
pilhuhn: I think I would expect that those 15/30/45 would scale with the max and be at 1/4, 1/2 and 3/4 of the max

burmanm: Do we still have those UI pages where the scrolling starts to be buggy if there's more than X items per page? I'm not sure if all of those were fixed..

josejulio: Uhg, I'm not sure. I might have to test with a ton of items.
I recall seeing in the code a bug when we had lots of pages. Not sure if you are refering to that one. https://github.com/josejulio/rhq/blob/d5dfe2ed9d42af51c6c3d3793f79e315c52ed092/modules/enterprise/gui/portal-war/src/main/webapp/rhq/resource/include/pagination-plain.xhtml#L66-L66

PR #287, opened 2017-01-16
Bug 1386913 - Remove old version from package_bits when deploys a new version

rubenvp8510:


burmanm:
This does not remove anything without unlinking, that's a problem (DELETE does not delete the object)

rubenvp8510:
@burmanm  I know, this is not ready yet.. but that's the idea.

rubenvp8510:
@burmanm @josejulio Could you please review ?

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -626,6 +627,59 @@ public void completeDeployPackageRequest(DeployPackagesResponse response) {
...
+            packageVersionQuery.setParameter("packageTypeName", key.getPackageTypeName());
+            packageVersionQuery.setParameter("architectureName", key.getArchitectureName());
+            packageVersionQuery.setParameter("version", key.getVersion());
+            packageVersionQuery.setParameter("resourceTypeId", resourceTypeId);
+            PackageVersion packageVersion = (PackageVersion) packageVersionQuery.getSingleResult();
+            purgePackageBits(packageVersion.getGeneralPackage().getId());
+        }
+
+        // Delete orphanate bits
+        if(DatabaseTypeFactory.isPostgres(DatabaseTypeFactory.getDefaultDatabaseType())) {
```
josejulio: Why only on Postgres ?

modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
```diff
@@ -290,6 +292,7 @@
     public static final String QUERY_FIND_BY_ID = "PackageVersion.findById";
     public static final String QUERY_FIND_PACKAGEVERSION_BY_FILENAME = "PackageVersion.findPackageVersionByFilename";
     public static final String QUERY_FIND_DELETEABLE_IDS_IN_REPO = "PackageVersion.findDeleteableVersionIds";
+    public static final String QUERY_FIND_PACKAGE_HISTORICAL_VERSIONS = "PackageVersion.findPackageOtherVersions";
```
josejulio: Could you please use the same name for both the constant and the string content? 
Unless I'm missing something we might prefer to be consistent.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
rubenvp8510: Becuase the unlink blob operation only applies on Postgres database.

modules/core/domain/src/main/java/org/rhq/core/domain/content/PackageVersion.java
rubenvp8510: Sure. I'll change it.

josejulio:
I suppose this only cleans the previous versions bits when deploying.
There should be a way to clean the ones that are already there for users who already have lots of these resources.

rubenvp8510:
@josejulio  I think you're right about it, I was thinking about defining a new operation for cleaning old content of the deployments.

WDYT? That sounds good? or it is better to have a script for that?

josejulio:
IMO the operation sounds good. 

I tested the current PR. When uploading the first time it creates a pg_largeobject, on the second time it creates a second pg_largeobject, successive uploads won't create any other pg_largeobject. 
Though the first pg_largeobject is successfully removed by running  `Remove orphan large objects` on `postgres`. I guess that this is more related to https://github.com/rhq-project/rhq/pull/286

loleary:
I do not think an operation would be the right way to go. Granted, such a capability might be nice on the PostgreSQL plug-in itself to remove orphaned pg_largeobject data, but that is really unrelated to what we are doing here, isn't?

Basically, what this fix should be doing is 2 things.

1. When a new version of content is deployed, the previous version should be removed. This is for content only such as `Deployment` or `DomainDeployment` resource types.
2. To clean up all the _old unreachable content_, a job should be executed that will remove it all asynchronously. Perhaps when a new version is deployed, a check should be performed to see if there are multiple old versions of the exact same content, and if so, schedule a job to remove them. I say _schedule_ because the process may take some time and we do not want that holding up the deployment of new content.

From my perspective, the operation becomes useless once we do 1.

loleary:
Also, I just want to make sure that proper discussion, or at least thought, has occurred surrounding the implications of actually deleting the old versions. You see, this _bug_ is actual a feature. The reason that it is broken is due to the UI being broken. In other words, the content pages are supposed to provide the user the ability to revert and remove old content. However, that feature of the UI seems to be using old JSF components and is now returning 404s. 

My point and concern is that if for any reason, a user demands that we fix the 404s that plague the content pages, we either need to undo this change or make certain that the _to-be-fixed_ content pages do not provide a revert or delete feature and instead, only provide the content history.

loleary:
And one last thing, is the clean-up of the old versions actually removing the package version entry? I think that would be a bad idea because package history would be lost and that is what users are using for auditing.

We should only be nulling the package bits ID in the package version table and deleting and unlinking the entry in package_bits and pg_largeobject.

i.e., all data remains intact except the BLOB is removed and the reference to the BLOB is removed.

rubenvp8510:
@loleary that is exactly what this PR does, it doesn't delete the package version entry, it just set package bits id to NULL and removes the package bits entry, ( also in the case of Postgres unlink the BLOB)

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -626,6 +628,76 @@ public void completeDeployPackageRequest(DeployPackagesResponse response) {
...
+        List<String> resourceTypes = Arrays.asList("Deployment", "DomainDeployment");
+        Query query = entityManager.createQuery("SELECT ip.packageVersion.generalPackage FROM InstalledPackage ip " +
+                "WHERE ip.resource.id IN (SELECT t.id FROM Resource t WHERE t.resourceType.id IN " +
+                "(SELECT id FROM ResourceType WHERE name IN ( :resourceTypes) ) )");
+        query.setParameter("resourceTypes", resourceTypes);
+        List <Package> packages = query.getResultList();
+        for(Package pkg: packages){
+            purgePackageBits(pkg.getId());
+        }
+        deleteOrpahanteBlobs();
```
josejulio: delete Orp**haned** Blobs

josejulio:
@loleary Rubén exposed a method to delete all the old history bits. Its meant to be called from a CLI script on servers that already have lots of history bits.
WDYT?

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
rubenvp8510: Hehehe, ups.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -626,6 +628,76 @@ public void completeDeployPackageRequest(DeployPackagesResponse response) {
...
+            packageVersionQuery.setParameter("version", key.getVersion());
+            packageVersionQuery.setParameter("resourceTypeId", resourceTypeId);
+            PackageVersion packageVersion = (PackageVersion) packageVersionQuery.getSingleResult();
+            purgePackageBits(packageVersion.getGeneralPackage().getId());
+        }
+
+        deleteOrphanedBlobs();
+    }
+
+    public void removeHistoryDeploymentsBits(){
```
loleary: Shouldn't all public API calls contain JavaDoc? Otherwise, this will be one more method that shows up in our product documentation without any information on the purpose or how it works.

loleary: I feel this method should require authentication. In its current form, anyone is able to remove the deployment bits.

Although this is probably fine for this release as this data is unreachable and unusable, my concern is if we change this in the future. It seems that there should be some protection for this method that requires 

Also not sure how easy the permission check will be. Perhaps this will be a super-user function? Otherwise, the user will have to have the manage resource permission for the resource the history is being removed from.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerRemote.java
```diff
@@ -291,4 +291,7 @@ PackageTypeAndVersionFormatComposite findPackageTypeWithVersionFormat(Subject su
      */
     PackageVersion createPackageVersionWithDisplayVersion(Subject subject, String packageName, int packageTypeId,
         String version, String displayVersion, Integer architectureId, String temporaryContentHandle);
+
+
+    void removeHistoryDeploymentsBits();
```
loleary: As stated in the removeHistoryDeploymentsBits() impl, shouldn't this require authentication (Subject)?

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
josejulio: I suppose this should be a super-user function, as is only needed for cleaning old packages and not a daily thing to do.

rubenvp8510:
I made some changes on the queries to be safer, I haven't tested it yet, I'll test it and If no ones have any other observation, we can merge.

rubenvp8510:
Also added permission check to `removeHistoryDeploymentsBits` function

rubenvp8510:
I changed a little bit the query , now it filters by plugin and resourceType, @loleary Could you please review the query? Thanks.

rubenvp8510:
This PR also removes the bits of old versions for packages when the user deploys a new version.
I think we can remove that behavior and leave only the CLI method, that would be safer because
only users that need to purge the versions will run the method

@josejulio  @loleary  WDYT?

loleary:
The updated query in 6b2f422 looks fine to me.

The CLI method is only intended as a general fix for environments that have been affected by this problem. The other part of the PR that removes the package history for a specific package upon update should remain as it is the actual fix for the original issue. 

The expectation going forward is:

1. User deploys version 1 of of HelloWorld.war.
2. User updates HelloWorld.war to version 2.
3. User updates HelloWorld.war to version 3.

The package history reflects version 1, 2, and 3 along with the deploy date, status, etc. but the actual BLOB for version 1 and 2 are no longer in the database. This is because when version 2 was deployed, version 1's BLOB was removed. And when version 3 was deployed version 2's BLOB was removed.

The only exception is if version 1 and version 2 are in a package repository, in which case, they will remain as they are linked to the repository.

However, if the above scenario was performed prior to this fix and then:

4. Apply update/fix
5. User updates HelloWorld.war to version 4.

Before step 5. is performed, the BLOBs for version 1, 2, and 3 are all in the database. After step 5., the BLOBs for version 1, 2, and 3 are removed.

Finally, the CLI method can be used in cases where steps 1 through 3 have been performed and there is no plan for a new version to be deployed. In that case, rhqadmin could use the CLI operation to remove all prior BLOBs from the database.

josejulio:
Did some testings and since @loleary is OK with the changes I'm going to merge it.

PR #288, opened 2017-01-16
Bug 1368107 - Update the minimumPagesize to match the new minimum.

josejulio:
Increases the amount on PagedDataTable to match PageControlSettingsUIBean.

This is a follow up on https://github.com/rhq-project/rhq/pull/285.
When modifying 15/30/45 to scale with the max I forgot to also scale the minimum value.
Also noticed that there was another minimum page that was being used for listing the rows.


josejulio:
@rubenvp8510 , can you take a look a this?

PR #290, opened 2017-01-31
[BZ 1272328] Ability to decommission maintenance mode storagenode'

burmanm:
Give user the ability to force storage node out of maintenance mode as well as change some behavior for decommission and maintenance mode flag.


jsanda:
I think the changes look good. It puts it in the hands of the user, but under the circumstances, I think that is ok.

PR #291, opened 2017-02-02
WIP Bug 1386913 - Newly uploaded content results in previous content vers…

rubenvp8510:
…ion being left in database with no way to clean it up

Needs more testing, don't merge yet... @josejulio


jshaughn:
If I understand correctly, this fix may increased the execution time of updateBackingContent and therefore increase the delay before retrieveBackingContent can be successfully called.

The removal of unwanted, older, package versions seems to me more like an async/purge type operation rather than an inline removal like we have here.  I'd suggest perhaps looking at a purge approach that just looks for older package versions and wipes them out.

PR #292, opened 2017-02-03
Prevents NPE on mergeExistingResourcePluginConfiguration when property definition doesn't exist anymore.

rubenvp8510:
@josejulio please review.

### DIFF
--- a/modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
+++ b/modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
@@ -1945,7 +1945,8 @@ public class InventoryManager extends AgentService implements ContainerService, 
             Property discoveredProp = pluginConfig.get(propertyName);
             // Only overrides the property if it is a read-only property
             // See BZ: 1379834
-            if (propDefs.get(propertyName).isReadOnly()) {
+            PropertyDefinition propDefinition = propDefs.get(propertyName);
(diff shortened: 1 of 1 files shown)


modules/core/plugin-container/src/main/java/org/rhq/core/pc/inventory/InventoryManager.java
```diff
@@ -1945,7 +1945,8 @@ private Configuration mergeExistingResourcePluginConfiguration(Resource resource
             Property discoveredProp = pluginConfig.get(propertyName);
             // Only overrides the property if it is a read-only property
             // See BZ: 1379834
-            if (propDefs.get(propertyName).isReadOnly()) {
+            PropertyDefinition propDefinition = propDefs.get(propertyName);
```
josejulio: Maybe a bit of paranoia, but, is `propDefs` always non `null` ?
https://github.com/rhq-project/rhq/pull/273/files#diff-32f6b984f4c7846e751c48497839e07aR1938

burmanm: It's never null. 

https://github.com/rhq-project/rhq/blob/master/modules/core/domain/src/main/java/org/rhq/core/domain/configuration/definition/ConfigurationDefinition.java#L182

josejulio: Thanks.

josejulio:
LGTM

PR #293, opened 2017-02-05
Bug 1386913 - Newly uploaded content results in previous content vers…

rubenvp8510:
…ion being left in database with no way to clean it up

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
@@ -123,6 +123,7 @@ import org.rhq.enterprise.server.plugin.pc.content.PackageTypeBehavior;
 import org.rhq.enterprise.server.resource.ResourceManagerLocal;
 import org.rhq.enterprise.server.resource.ResourceTypeManagerLocal;
 import org.rhq.enterprise.server.resource.ResourceTypeNotFoundException;
+import org.rhq.enterprise.server.scheduler.jobs.DataPurgeJob;
 import org.rhq.enterprise.server.util.CriteriaQueryGenerator;
 import org.rhq.enterprise.server.util.CriteriaQueryRunner;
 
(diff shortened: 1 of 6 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -223,6 +224,12 @@ public void mergeDiscoveredPackages(ContentDiscoveryReport report) {
             LOG.debug("Finished merging [" + report.getDeployedPackages().size() + "] packages in "
                 + (System.currentTimeMillis() - start) + "ms");
         }
+
+        try{
+            DataPurgeJob.purgeNow();
```
josejulio: I guess that is OK for quick testing, but we might prefer to wait for the DataPurgeJob to be executed. AFAIK it executes every hour or so by default.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -630,32 +637,10 @@ public void completeDeployPackageRequest(DeployPackagesResponse response) {
...
-            PackageVersion packageVersion = (PackageVersion) packageVersionQuery.getSingleResult();
-            purgePackageBits(packageVersion.getGeneralPackage().getId());
-        }
-
-        deleteOrphanedBlobs();
     }
 
-    public void removeHistoryDeploymentsBits(Subject user){
-
-        if (!authorizationManager.isSystemSuperuser(user)) {
```
josejulio: Any reason to remove this?
I think we should bring this back.

josejulio: Oh ok, it seems like this unprotected method is only on the local API.

josejulio:
Tested it using GUI and CLI and I didn't see any issue. Still I suggest to remove the call to `purgeNow` to avoid overloading the server with frequent calls to the `PurgeJob`.

josejulio:
Actually we should remove that purgeNow call. 
It is being called on every discovery.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/scheduler/jobs/DataPurgeJob.java
```diff
@@ -145,6 +146,12 @@ private void purgeEverything(SystemSettings systemSettings) {
         purgeResourceConfigHistory(systemSettings);
         removeResourceErrorDuplicates();
         removeStaleAvailabilityResourceErrors();
+        purgeOldPackageBits();
+    }
+
+    private void purgeOldPackageBits(){
+        ContentManagerLocal cm = LookupUtil.getContentManager();
```
burmanm: Rest of the class does this lookup in the DataPurgeJob instantiation, this should be there also for the clarity.

josejulio:
@rubenvp8510 
The purgeJob is scheduled here: https://github.com/josejulio/rhq/blob/7fb9222c80981fb876d8a7eea472304761f42555/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/core/StartupBean.java#L776-L778
`every hour, on the half-hour`

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerBean.java
```diff
@@ -671,47 +650,56 @@ public void removeHistoryDeploymentsBits(Subject user){
...
-    private void deleteOrphanedBlobs(){
-        // Delete orphanate bits
-        if(DatabaseTypeFactory.isPostgres(DatabaseTypeFactory.getDefaultDatabaseType())) {
-            try {
-                Connection conn = dataSource.getConnection();
-                PreparedStatement ps =
-                        conn.prepareStatement("SELECT BITS FROM " + PackageBits.TABLE_NAME + " WHERE ID NOT IN ( " +
-                                "SELECT PACKAGE_BITS_ID FROM RHQ_PACKAGE_VERSION WHERE PACKAGE_BITS_ID IS NOT NULL )");
-                ResultSet rs = ps.executeQuery();
+    public void removeHistoryDeploymentsBits(Subject user){
```
josejulio: Since is now on the purgeJob, it might be worth to remove this public API (one thing less to document).

josejulio:
Testing

josejulio:
Did some deployments till I had some orphaned rhq_package_bits.
After the PurgeJob executed i had 0 orphaned rhq_package_bits or pg_largeobjects.

rubenvp8510:
So, I think this PR is ready for merge, isn't it? :)

PR #294, opened 2017-02-09
BZ-1386913 Separate transactions for remove bits.

rubenvp8510:
- Added validations for unlink only on Postgres
- Put unlink process as part of other transaction
- One transaction per package.

@josejulio Could you please help me to test and review? Thank you!!


josejulio:
It looks good, will test.

PR #295, opened 2017-03-01
Bug 1417868 - Missing "per minute" web-metrics for EAP 7

rubenvp8510:


rubenvp8510:
@josejulio Could you please review?

PR #296, opened 2017-03-04
Bug 1427571 - Solve Unusable group metric graphs.

josejulio:
Removes the black circle and non-existent data, adds a dot to mark existent data.

Before:
graph01_before

After:
graph01_after

Before:
graph02_before

After:
graph02_after

### DIFF
--- a/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/groups/detail/monitoring/table/CompositeGroupD3MultiLineGraph.java
+++ b/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/groups/detail/monitoring/table/CompositeGroupD3MultiLineGraph.java
@@ -291,7 +291,11 @@ public class CompositeGroupD3MultiLineGraph extends CompositeGroupD3GraphListVie
                         .attr("stroke", function(d,i){ return colorScale(i);})
                         .attr("stroke-width", "2")
                         .attr("stroke-opacity", ".9")
-                        .attr("d", function(d) { return graphLine(d.value);});
+                        .attr("d", function(d) { return graphLine(
+                            d.value.filter(function(d) {
+                                return ('nodata' in d)? !d.nodata : true;
+                            })
+                        );});
 
                 for (var i=0;i<chartContext.data.length;++i) {
                     svg.selectAll("dot")
@@ -301,11 +305,11 @@ public class CompositeGroupD3MultiLineGraph extends CompositeGroupD3GraphListVie
                         .filter(function(d) {
                             return ('nodata' in d)? !d.nodata : true;
                         })
-                        .attr("stroke", "black")
-                        .attr("stroke-width", "2")
+                        .attr("stroke", function(){ return colorScale(i);})
+                        .attr("stroke-width", "1")
                         .attr("stroke-opacity", ".9")
-                        .attr("fill", "none")
-                        .attr("r", 3.5)
+                        .attr("fill", function(){ return colorScale(i);})
+                        .attr("r", 2)
                         .attr("cx", function(d) {
                             return timeScale(d.x);
                         })


burmanm:
@Pathduck Do these look better for your use-case?

Pathduck:
@burmanm Yes indeed they do - they look very good and I think the dots make it easier to see data points - as long as they are small and of the same colour as the line it should not be a problem :)

burmanm:
My only comment really is the lowest image, how come the filling does not start from the graph's left side (from unvisible datapoint) ?

Pathduck:
I agree with that last point, it should start from the leftmost side.

loleary:
> My only comment really is the lowest image, how come the filling does not start from the graph's left side (from unvisible datapoint) ?

I agree that it would be ideal to see the line begin at the left of the chart (and also continue to the right). However, I am not sure how this will really look and if it may confuse things more.

Basically, when I originally thought about this I could only come up with the following solutions:

1. The line will start at the first data point in the series.
2. Just start the line at the left of the graph using the value of the first data point. This will result in a straight level line from the left edge to at least the first data point.
3. A dotted or dashed line would be used when no data is available and we would use the value of the first data point to draw a straight dotted line from the left edge to the first data point.
4. Load enough data for the specific metric to cover the next available data that falls before and after the chart's time range. This can then be used to plot a proper line with appropriate curve to represent the previous and next known data points as they relate to the first and last available data points in the chart's time range.

For the proposed PR josejulio has chosen 1. This results in no markings or graph line before the first available data point or after the last available data point.

Thought 2. doesn't seem valid to me as it would misrepresent the value for the period leading into and out of the chart's time range. However, with the dots representing real data values, maybe this isn't bad. It just seems like the next bug is going to be something like, "Why is my CPU time reported at 1.75% at 5:05 PM when my graph is covering 5:05 to 6PM but it is reported at 0.5% at 5.05 PM when I set my time period to 4:10 PM to 5:05 PM?"

Thought 3. is what I originally suggested as a possible solution but the fear was that it would make the chart unreadable. If 3. was used, it would mean the areas where we do not see a line in the example screen shots would be filled in with a straight dotted/dashed line from the left edge to the first data point and from the last data point to the right edge using the value from the respective real data point. However, when I proposed this solution I also thought it better to use the same dotted/dashed line for the time slices that fall between the real data values.

Thought 4. would be the nicest but would require a dramatic change in how metric data is loaded for charting and would require additional calculations to product the averaging curve for the data points outside of the chart's time range while also applying the appropriate time scale to the curve.

I therefore think option 1. makes the most sense unless there is some benefit to the upstream RHQ Project.

josejulio:
Option 1 makes the most sense to me too, not only because is already done, but it would be less confusing to the user.

Pathduck:
I think the original graphs "solved" this (not very elegantly) by just drawing from 0 at the start. At least that's what it looks like for the old examples.

I think if the start of the X axis starts at 5:05 then it should start with the data value collected at 5:05, if any. Might be some deeper reasons why that won't work though...maybe there was no data collected at that point. Or, avoid drawing a time stamp at the Y axis point somehow.

The very old type graphs had the same issue as well, so not sure how big of a problem it would be:
https://developer.jboss.org/message/915291#915291

josejulio:
> I think if the start of the X axis starts at 5:05 then it should start with the data value collected at 5:05, if any. Might be some deeper reasons why that won't work though...maybe there was no data collected at that point. 

That's the current case, there is no data. I didn't wanted to set it on zero because it would lead to believe that previous data might be 0 or something like that.

> Or, avoid drawing a time stamp at the Y axis point somehow.

I think that if we avoid drawing the timestamp where there was no collection intervals (on all the displayed graphs) will lead to confusions like, "I want my graph to cover 5:05PM to 6PM but is only showing 5:30PM to 6:PM"

Pathduck:
> I think that if we avoid drawing the timestamp where there was no collection intervals (on all the displayed graphs) will lead to confusions like, "I want my graph to cover 5:05PM to 6PM but is only showing 5:30PM to 6:PM"

@josejulio Sorry, misunderstanding :)

What I meant was just not draw a time stamp at start of X axis (where Y axis is), in this instance do not draw 05:05 - not for every point, that would indeed be confusing :)

josejulio:
I'm sure that the 5:05 will be hidden. 
I think it is being show on the 'after' graphs because I'm setting the interval manually.

josejulio:
I think this might look a bit weird when we don't have any data, but once data is collected it should look good.
Might be more complicated to make the graph look "pretty" that its worth.

@burmanm  if you don't have any objections, please merge.

PR #297, opened 2017-03-08
Bug 1326707 - Cleaned relative-to option values

josejulio:


josejulio:
@burmanm Please review

PR #298, opened 2017-03-08
Bug 1427439 Adding a default transport to cache container.

rubenvp8510:


burmanm:
Is this really affecting the correct place? Now you're modifying the createResource of the whole Infinispan instead of the add of "Cache" (which is still using TemplatedComponent).

I thought the BZ was talking about an issue surrounding the addition of say invalidation-cache, not Cache Containers / Infinispan Subsystem module.

rubenvp8510:
It is necessary for the case of distributed caches to have one transport created on the container, that is why EAP fails when you tried to create a cache inside one of the cache containers.

What I'm doing here is create a default jgroups transport.


You can see the discussion here: https://issues.jboss.org/browse/JBEAP-8667

josejulio:
Shouldn't we allow to specify the transport instead of setting a fixed one?
Also, if we don't yet allow transport creation, it should be exposed for this to work.
WDYT?

spinder:
Also, how does this operation work on EAP 7.0.x systems now that we're upgrading the "createResource" behavior for all infinispan 7.0.x and 7.1. managed resources?

spinder:
Feels like the cleaner thing to do here would be to modify the BaseComponent to: 
i)detect if 7.1.x api level, and 
ii)add new conditional createResource operation(jgroups,etc.) behavior when needed.

that way we don't risk breaking existing EAP 7.0.x servers and only add the new requirements when it's needed for 7.1.x.  Remember that the EAP plugin pack is reused by third party plugin teams, like JDG/FUSE/etc., so non-trivial changes here could have unforseen consequences. 

@rubenvp8510, how long would it take to make those changes?

josejulio:
According to the JIRA Rubén referenced,  it will fail on deployments that
use this invalid cached (on eap7).
Is required to have a transport, even on eap7, but wasn't enforce at this
point. So the right thing would be to fail if no transport is specified.

El El jue, 9 de marzo de 2017 a la(s) 15:18, spinder <
notifications@github.com> escribió:

Also, how does this operation work on EAP 7.0.x systems now that we're
upgrading the "createResource" behavior for all infinispan 7.0.x and 7.1.
managed resources?

—
You are receiving this because you commented.


Reply to this email directly, view it on GitHub
<https://github.com/rhq-project/rhq/pull/298#issuecomment-285486063>, or mute
the thread

spinder:
@josejulio can you say a bit more about what you mean by validate?  I'm still a little concerned adding this new 'correct' configuration component for EAP 7.0.x and potentially break something in JDG.  What if they already work around or properly fix this infinispan configuration issue?  Specifying the right default transport for 7.1.x managed resources definitely makes sense since there is not chance of breaking existing usage pattern.

josejulio:
> @josejulio can you say a bit more about what you mean by validate?

I was thinking on also failing on EAP 7.0 when no `transport` is specified.

> I'm still a little concerned adding this new 'correct' configuration component for EAP 7.0.x and potentially break something in JDG. What if they already work around or properly fix this infinispan configuration issue? Specifying the right default transport for 7.1.x managed resources definitely makes sense since there is not chance of breaking existing usage pattern.

If they already do a workaround for it then it should be the same for EAP7.1.

Anyway I get your point about trying to do the minimal impact possible. I guess that Ruben already knows of a way to detect EAP 7.1 (API level or something).

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/BaseServerComponent.java
```diff
@@ -1135,4 +1138,23 @@ protected String collectPatches() {
             return ret.toString();
         }
     }
+
+    Integer getMinorVersion(){
+        String version = serverPluginConfig.getVersion();
```
burmanm: What's the point of this? We can already detect EAP 7.0 / EAP 7.1 difference by looking at the API version. 

https://github.com/rhq-project/rhq/blob/master/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/JBossProductType.java#L96

Lets not do multiple types of detection.

rubenvp8510: You're right, I changed the detection type, now I'm using the api version.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/InfinispanComponent.java
```diff
@@ -0,0 +1,75 @@
...
+                if (result.isSuccess()) {
+                    report.setStatus(CreateResourceStatus.SUCCESS);
+                    report.setResourceKey(createAddressCacheContainer.getPath());
+                    report.setResourceName(report.getUserSpecifiedResourceName());
+                } else {
+                    report.setStatus(CreateResourceStatus.FAILURE);
+                    report.setErrorMessage(result.getFailureDescription());
+                }
+                return report;
+            } else {
```
burmanm: Nitpicking, but these two else parts are useless, you could have just replaced them with one return call ;)

PR #301, opened 2017-03-28
Bug 1433904 - Lists IO subsystem and show options for worker and buffer-pool forms

josejulio:
It currently loads all the buffer-pool's and worker's on the http, https, ajps, etc. So you can choose one.
screenshot from 2017-03-28 14-30-58

But it has one (know) issue. While it loads only the needed items (buffer-pools or workers) on each standalone, it mixes all items on profile mode, because the filtering is done at the child after the `PLATFORM`  (see [here](https://github.com/josejulio/rhq/blob/8ae363d095c697a7521fb9ba17a6432ff34d4532/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java#L2854))

Only solution i can think on would be to modify a bit the XSD to extend expressionScope and include a expresionScopeQuery to filter on the type=Profile. 

Alternatives are:

- Modify the XSD (needs to modify the core).
- Put another separator on the [expression](https://github.com/rhq-project/rhq/pull/301/files#diff-6b0d9c8ba3ea36d420fa47beb16cac2eR6829) to make room for a query for baseResource (needs to modify the core).
- Leave it as is, it will fail if you chose a pool that isn't on that Profile.
- Drop the auto filling  on Profile.
- Do it on a different way.

I'm sure @burmanm would know something else.

### DIFF
--- a/modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
+++ b/modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
@@ -3317,6 +3317,48 @@
       </service><!-- End of servlet-container -->  
     </service><!-- End of undertow service -->
 
+    <service name="IO (Managed Server)"
+             discovery="SubsystemDiscovery"
+             class="BaseComponent"
+             singleton="true"
+             subCategory="Subsystems|Web">
+
+        <plugin-configuration>
+            <c:simple-property name="path" readOnly="true" default="subsystem=io"/>
+            <c:simple-property name="managedRuntime" default="true" type="boolean" readOnly="true"/>
+        </plugin-configuration>
+
+        <service name="Worker (Managed Server)"
+                 discovery="SubsystemDiscovery"
+                 class="BaseComponent"
+                 createDeletePolicy="neither">
+            <plugin-configuration>
+                <c:simple-property name="path" readOnly="true" default="worker"/>
+            </plugin-configuration>
+
+            <resource-configuration>
+                <c:simple-property name="io-threads" required="false" type="integer" readOnly="true" description="Specify the number of I/O threads to create for the worker.  If not specified, a default will be chosen, which is calculated by cpuCount * 2"/>
+                <c:simple-property name="stack-size" required="false" type="long" readOnly="true" defaultValue="0" description="The stack size (in bytes) to attempt to use for worker threads. The default value is 0."/>
+                <c:simple-property name="task-keepalive" required="false" type="integer" readOnly="true" defaultValue="60" description="Specify the number of milliseconds to keep non&#45;core task threads alive. The default value is 60."/>
+                <c:simple-property name="task-max-threads" required="false" type="integer" readOnly="true" description="Specify the maximum number of threads for the worker task thread pool.If not set, default value used which is calculated by formula cpuCount * 16"/>
+            </resource-configuration>
+        </service>
+
+        <service name="Buffer Pool (Managed Server)"
+                 discovery="SubsystemDiscovery"
+                 class="BaseComponent"
+                 createDeletePolicy="neither">
+            <plugin-configuration>
+                <c:simple-property name="path" readOnly="true" default="buffer-pool"/>
+            </plugin-configuration>
+
+            <resource-configuration>
+                <c:simple-property name="buffer-size:expr" required="false" type="string" readOnly="true" description="The size of each buffer slice in bytes, if not set optimal value is calculated based on available RAM resources in your system."/>
+                <c:simple-property name="buffers-per-slice:expr" required="false" type="string" readOnly="true" description="How many buffers per slice, if not set optimal value is calculated based on available RAM resources in your system."/>
+                <c:simple-property name="direct-buffers" required="false" type="boolean" readOnly="true" description="Does the buffer pool use direct buffers, some platforms don&apos;t support direct buffers"/>
+            </resource-configuration>
+        </service>
+    </service>
     <service name="JCA (Managed Server)"
              class="BaseComponent"
              discovery="SubsystemDiscovery"
@@ -6783,7 +6825,9 @@
             <c:simple-property name="security-key" required="false" type="string" readOnly="false" description="The security key that is used for the mod&#45;cluster group. All members must use the same security key."/>
             <c:simple-property name="security-realm" required="false" type="string" readOnly="false" description="The security realm that provides the SSL configuration"/>
             <c:simple-property name="use-alias" required="false" type="boolean" readOnly="false" defaultValue="false" description="If an alias check is performed. The default value is false."/>
-            <c:simple-property name="worker" required="false" type="string" readOnly="false" defaultValue="default" description="The XNIO worker that is used to send the advertise notifications. The default value is default."/>
+            <c:simple-property name="worker" required="false" type="string" readOnly="false" defaultValue="default" description="The XNIO worker that is used to send the advertise notifications. The default value is default.">
+                <c:option-source target="resource" expressionScope="baseResource" expression="type=Worker plugin=&pluginName;"/>
+            </c:simple-property>
           </resource-configuration>
 
         </service><!-- End of Mod cluster service -->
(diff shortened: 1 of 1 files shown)


loleary:
> * Modify the XSD (needs to modify the core).

When you say modify core, what do we mean? I am assuming you mean outside of the plug-in?

> * Put another separator on the expression to make room for a query for baseResource (needs to modify the core).

When you say modify core, what do we mean? I am assuming you mean outside of the plug-in?

Is there a reason that this would be worse then updating the XSD? Obviously, overloading a specific function with too many meanings is bad but I would like to understand the difference between adding a `expresionScopeQuery` attribute and extending/overloading the function of `expression`. 

In either case, I am guessing you plan to limit the source to the resources that are defined under the same _Profile_?

> * Leave it as is, it will fail if you chose a pool that isn't on that Profile.

I do not think this will work. As it stands, wouldn't you end up with multiple `default` resources assuming there are multiple _Profiles_? How would the user determine which one would be applicable in their specific case?

> * Drop the auto filling on Profile.

Do you mean that these values would become property strings? The user would need to manually enter the appropriate resource name for the IO Worker or IO Buffer Pool?

> * Do it on a different way.

I am not sure there would be a different way. But that's just me.

josejulio:
> When you say modify core, what do we mean? I am assuming you mean outside of the plug-in?

Yes, will need to modify `rhq-core-client-api` and `rhq-enterprise-server`.

> Is there a reason that this would be worse then updating the XSD? Obviously, overloading a specific function with too many meanings is bad but I would like to understand the difference between adding a expresionScopeQuery attribute and extending/overloading the function of expression.

I'm just enumerating some options that came to my mind. I would prefer to update the XSD and add the optional attribute `expresionScopeQuery` to avoid having to do funny things with the `expression` attribute while being backwards compatible with existing plugins.

> In either case, I am guessing you plan to limit the source to the resources that are defined under the same Profile?

Yes, I would like to narrow it to the current parent Profile.

> I do not think this will work. As it stands, wouldn't you end up with multiple default resources assuming there are multiple Profiles? How would the user determine which one would be applicable in their specific case?

The attribute excepts a string, any `default` would work. Anyway this currently strips duplicates.
profile


> Do you mean that these values would become property strings? The user would need to manually enter the appropriate resource name for the IO Worker or IO Buffer Pool?

That's correct.

> I am not sure there would be a different way. But that's just me.

I was actually hoping @burmanm would know other options.

burmanm:
I'm not sure if there's another option.. I remember that we've had this same issue at some point in the past also, but I can't find the relevant BZ / commits.

burmanm:
@loleary I'm wondering from some past issues, are these really settings that we want to modify under (Profile) and not just under (Managed) ? Don't they replicate if you add them from there?

Second approach I could think of is overriding the loadResourceConfiguration() and then do if(parentResource instanceof ..) and do the filtering manually there for available pools. That we way we wouldn't have to change the XSD and this can be done in the plugin-patch only.

josejulio:
> Second approach I could think of is overriding the loadResourceConfiguration() and then do if(parentResource instanceof ..) and do the filtering manually there for available pools. That way we wouldn't have to change the XSD and this can be done in the plugin-patch only.

I'll take a look at it. I thinking on doing that from the beginning, to load the needed resources, but couldn't see any way to create the option list. I'll take a more in-deep look at that. Thanks.

josejulio:
I don't think I can filter this on `loadResourceConfiguration`.
I actually want to change some attributes on ConfigurationDefinition, not the Configuration itself.

Furthermore, the options are being resolved on a separate function, called by [GWT](https://github.com/josejulio/rhq/blob/b6cdc1e407ad4b5eff5be3d258beb7fae9fe8c05/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/server/gwt/ConfigurationGWTServiceImpl.java#L377-L377).

josejulio:
> Is there a reason that this would be worse then updating the XSD? Obviously, overloading a specific function with too many meanings is bad but I would like to understand the difference between adding a expresionScopeQuery attribute and extending/overloading the function of expression.

Forgot to mention that updating the XSD will likely require a new colum on the table [RHQ_CONFIG_PD_OSRC](https://github.com/josejulio/rhq/blob/07bead42ccd636be67b717e9922f5a4eb134cff8/modules/core/domain/src/main/java/org/rhq/core/domain/configuration/definition/PropertyOptionsSource.java#L60-L60)

josejulio:
I "overloaded" the expression string to include a token `;` to use a different base resource.
It walks searching an ancestor of the type defined after the ';' e.g. `Profile`.
If one is found is used.

This avoid modifying the XSD and thus requiring a new column.

spinder:
@josejulio, if you need to modify the ConfigurationDefinition, why not create a new component and delegate(something like ConfigurationReadWriteDelegate)?  Since this is our first time introducing the io/subsystem we can extend and change behavior here without worrying about affecting any existing customers.  The more we modify the core, the more complicated and risky the fix delivery becomes. Also, we might want to reach out to @pilhuhn to see if he has any additional/different suggestions here that are more contained.

josejulio:
@spinder I could find any way to do that (I'm missing something?).

I need to hand over the ConfigurationDefinition (with the worker/buffer pool options) to GWT and haven't found (yet) a way to do that using only plugin-code.

The loadResourceConfiguration won't help much in that case, as the Configuration does not have the notion of options. For it is just a SimpleProperty with a string value, on tthe ConfigurationDefinition you can specify which options are rendered in the GUI. In case the options are in a  `option-source` (a kind of datasource) it asks to the server to [resolve](https://github.com/josejulio/rhq/blob/b6cdc1e407ad4b5eff5be3d258beb7fae9fe8c05/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/server/gwt/ConfigurationGWTServiceImpl.java#L377-L377) those values. I'm supposing that modifying the ConfigurationDefinition from within the loadResourceConfiguration won't do any good, because i *think* is shared  (I'll look more onto it, if is not shared and that definition is handed over to the UI, then that's the way to go) and i'm not really sure if that one is called on the server side or the agent.

So if I could just pass the UI a ConfigurationDefinition (using plugin code) with the values already there that would be good. Else i'm stuck doing this :(
I'll try to find something else.

burmanm:
@josejulio ConfigurationLoadDelegate can load the ConfigurationDefinition also (and write it). So that's why @spinder recommended it.

Technically you can also just request the available values from the server itself, like BaseComponent does. So you could do ASConnection in the loadResourceConfiguration, iterate through the values and only leave the matching ones.

josejulio:
@burmanm @spinder 
Those options would fit well for `map-property` or `list-property`. In this case, I need a `simple-property` with options to choose from. Is either a fixed number of options:

```xml
<c:simple-property name="unit" type="string" readOnly="true" description="The time unit">
    <c:property-options>
        <c:option value="NANOSECONDS"/>
        <c:option value="MILLISECONDS"/>
        <c:option value="SECONDS"/>
        <c:option value="MINUTES"/>
        <c:option value="HOURS"/>
        <c:option value="DAYS"/>
    </c:property-options>
</c:simple-property>
```
 or a datasource (like the one i'm using):
```xml
<c:simple-property name="workmanager" required="false" type="string" readOnly="true" description="The WorkManager instance for the BootstrapContext">
    <c:option-source target="resource" expression="type=Workmanager plugin=&pluginName;"/>
</c:simple-property>
```

> Technically you can also just request the available values from the server itself, like BaseComponent does. So you could do ASConnection in the loadResourceConfiguration, iterate through the values and only leave the matching ones.

Yep, but I haven't found anywhere to put those values. Let me clarify it a bit more.

The workflow is like this:

1. UI [requests](https://github.com/josejulio/rhq/blob/2eb88800a391e2d4681fc7a747223e4ce0039582/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/configuration/ResourceConfigurationEditView.java#L130-L130) latest `ResourceConfigurationUpdate`.
2. UI [requests](https://github.com/josejulio/rhq/blob/2eb88800a391e2d4681fc7a747223e4ce0039582/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/configuration/ResourceConfigurationEditView.java#L140-L140) from the cache `ResourceConfigurationDefinition` (or the DB).
3. UI [requests](https://github.com/josejulio/rhq/blob/2eb88800a391e2d4681fc7a747223e4ce0039582/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/configuration/ResourceConfigurationEditView.java#L145-L145) the `OptionValues` for the `ConfigurationDefinition`. This is what my latest changes are modifying.
4. With that, it builds the form.

Even if i used a `Component` or `Delegate`, I couldn't modify the options inside the `ConfigurationDefinition`(Which are fetched from the server/db) that are used to build the form.
To further clarify, I haven't found any way to let the UI know that I want to filter those values, as they aren't part of the Configuration.
On the component I get to use the `ConfigurationDefinition`, but I can't pass it back to the UI.

There is a [filter](https://github.com/josejulio/rhq/blob/2eb88800a391e2d4681fc7a747223e4ce0039582/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/configuration/ResourceConfigurationEditView.java#L159-L159) in that code, but doesn't seem to be any implementation anywhere of it.

Furthermore, this issue seems to be present on other subsystems (on profiles). I found at least this one in JCA:
screenshot from 2017-04-04 12-18-24

Notice how the workmanager has options for being `default`, `default_workmanager` (workmanager found on profile=default) and `full_ha_workmanager` (workmanager found on profile=full_ha.


Other options I can think of:

-   To revert latest changes and leave it like `Workmanager`
-   Create a new method on `BaseComponent`/`ConfigurationFacet` that does the filtering. Instead of overloading that `expression`


Thats all i could find, sorry if i'm not following something.

spinder:
@josejulio if indeed the issue that you're describing is not unique to the new "io" subystem then you should file an additional bug to capture that larger problem. It's possible that behavior has been affecting the EAP plugin for awhile now.

josejulio:
> spinder commented 15 minutes ago
> @josejulio if indeed the issue that you're describing is not unique to the new "io" subystem then you should file an additional bug to capture that larger problem. It's possible that behavior has been affecting the EAP plugin for awhile now.

Sure

josejulio:
I'm going to open a separate PR with commit https://github.com/rhq-project/rhq/pull/301/commits/0e9501faacbbac018927a286ae2670b8b258c048 and remove from this PR.

josejulio:
@spinder @loleary 
This PR only adds the IO subsystem, the other issue is being tracked on a different BZ/PR.
Is there anything else you would like on this PR?

spinder:
@josejulio from my side LGTM.  The other issue looks unreleated and most likely affecting current and previous EAP plugin versions as well. @loleary other thoughts?

loleary:
Looks fine to me.

josejulio:
@burmanm could you please review/merge?

PR #303, opened 2017-04-06
Bug 1439343 - Overload expression to accept a 'baseResource' type to …

josejulio:
…use instead of the topmost ancestor.

This code was originally on #301 but moved as a separate issue.

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
@@ -2655,17 +2655,24 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
 
             Predicate<Resource> baseResourcePredicate = null;
             if (expressionScope == PropertyOptionsSource.ExpressionScope.BASE_RESOURCE) {
-                int halfCommaIndex = expression.indexOf(";");
-                if (halfCommaIndex > -1) {
+                String[] expressionTokens = expression.split("\\s+");
+                String ancestorType = null;
+                final Pattern ancestorTypePattern = Pattern.compile("(.*);baseResource\\.plugin");
+
+                for (String token : expressionTokens) {
+                    Matcher ancestorTypeMatcher = ancestorTypePattern.matcher(token);
+                    if (ancestorTypeMatcher.lookingAt()) {
+                        ancestorType = ancestorTypeMatcher.group(1);
+                        break;
+                    }
+                }
+                if (ancestorType != null) {
                     if (resource == null && parentResource == null) {
                         LOG.warn("Different base resource type requested but resource id and parent_resource_id are not valid."
                                 + "Option source expression:" + expression);
                         return;
                     }
                     Resource nonNullResource = (resource == null? parentResource : resource);
-                    String ancestorType = expression.substring(
-                            halfCommaIndex + 1, expression.length());
-                    expression = expression.substring(0, halfCommaIndex);
                     Resource foundBaseResource =  ResourceUtility.getAncestorResourceOfType(
                             nonNullResource, ancestorType);
                     if (foundBaseResource != null) {
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
@@ -2657,7 +2657,7 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
             if (expressionScope == PropertyOptionsSource.ExpressionScope.BASE_RESOURCE) {
                 String[] expressionTokens = expression.split("\\s+");
                 String ancestorType = null;
-                final Pattern ancestorTypePattern = Pattern.compile("(.*);baseResource\\.plugin");
+                final Pattern ancestorTypePattern = Pattern.compile("(.*)BaseResource\\.plugin");
 
                 for (String token : expressionTokens) {
                     Matcher ancestorTypeMatcher = ancestorTypePattern.matcher(token);
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
@@ -2570,12 +2570,12 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
     public ConfigurationDefinition getOptionsForConfigurationDefinition(Subject subject, int resourceId,
         int parentResourceId, ConfigurationDefinition def) {
 
-        Resource resource = null, baseResource = null;
+        Resource resource = null, baseResource = null, parentResource = null;
         if (resourceId >= 0) {
             resource = resourceManager.getResource(subject, resourceId);
         }
         if (parentResourceId >= 0) {
-            Resource parentResource = resourceManager.getResource(subject, parentResourceId);
+            parentResource = resourceManager.getResource(subject, parentResourceId);
             baseResource = ResourceUtility.getBaseServerOrService(parentResource);
         } else if (resource != null) {
             baseResource = ResourceUtility.getBaseServerOrService(resource);
@@ -2586,19 +2586,19 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
 
             if (pd instanceof PropertyDefinitionSimple) {
                 PropertyDefinitionSimple pds = (PropertyDefinitionSimple) pd;
-                handlePDS(subject, resource, baseResource, pds);
+                handlePDS(subject, resource, parentResource, baseResource, pds);
 
             } else if (pd instanceof PropertyDefinitionList) {
                 PropertyDefinitionList pdl = (PropertyDefinitionList) pd;
                 PropertyDefinition memberDef = pdl.getMemberDefinition();
                 if (memberDef instanceof PropertyDefinitionSimple) {
                     PropertyDefinitionSimple pds = (PropertyDefinitionSimple) memberDef;
-                    handlePDS(subject, resource, baseResource, pds);
+                    handlePDS(subject, resource, parentResource, baseResource, pds);
                 } else if (memberDef instanceof PropertyDefinitionMap) {
                     PropertyDefinitionMap pdm = (PropertyDefinitionMap) memberDef;
                     for (PropertyDefinition inner : pdm.getOrderedPropertyDefinitions()) {
                         if (inner instanceof PropertyDefinitionSimple) {
-                            handlePDS(subject, resource, baseResource, (PropertyDefinitionSimple) inner);
+                            handlePDS(subject, resource, parentResource, baseResource, (PropertyDefinitionSimple) inner);
                         }
                         if (LOG.isDebugEnabled()) {
                             LOG.debug("3 ____[ " + inner.toString() + " in " + pdl.toString()
@@ -2616,7 +2616,7 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
                 PropertyDefinitionMap pdm = (PropertyDefinitionMap) pd;
                 for (PropertyDefinition inner : pdm.getOrderedPropertyDefinitions()) {
                     if (inner instanceof PropertyDefinitionSimple) {
-                        handlePDS(subject, resource, baseResource, (PropertyDefinitionSimple) inner);
+                        handlePDS(subject, resource, parentResource, baseResource, (PropertyDefinitionSimple) inner);
                     } else {
                         if (LOG.isDebugEnabled()) {
                             LOG.debug("4 ____[ " + inner.toString() + " in " + pdm.toString()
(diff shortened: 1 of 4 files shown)


josejulio:
I fixed some properties that could have some wrong options. (io subsystem is not yet here)

loleary:
@josejulio I spoke to @spinder regarding this PR. The issue I have is that it crosses two separate maintenance streams.
The changes to core server/domain are fine for this PR and BZ 1439343. But we will need a new BZ and PR for the changes needed to the plug-in(s).
Furthermore, we will need to think/discuss more about how this change will introduce dependencies on updates from separate maintenance streams.
For example, in the product stream the expected order of changes will be 3.3.9 followed by a new plug-in pack update version that provides the new expression syntax. However, for maintenance, we have to ensure that dependencies will not break environments that have not yet been updated. If I have NOT installed 3.3.9 I should still be able to get the latest plug-in pack update without anything breaking. Obviously, I would not get the fixes that the combined updates would provide but at least I would be no worse off.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
```diff
@@ -2652,6 +2653,34 @@ private void handlePDS(final Subject subject, Resource resource, Resource baseRe
             if (filter != null)
                 filterPattern = Pattern.compile(filter);
 
+            Predicate<Resource> baseResourcePredicate = null;
+            if (expressionScope == PropertyOptionsSource.ExpressionScope.BASE_RESOURCE) {
+                if (expression.contains(";")) {
```
burmanm: As expression.indexOf(";") is done so often in this codepart, wouldn't it make sense to do it once here?

```
int halfCommaIndex = expression.indexOf(";");
if(halfCommaIndex > -1) { 
.. processing code
}
```

Easier to read and change

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
```diff
@@ -2718,6 +2739,20 @@ private void handlePDS(final Subject subject, Resource resource, Resource baseRe
...
     }
 
+    private Iterable<Resource> processSearchExpression(String expression, ResourceCriteria criteria,
+                                                       CriteriaQueryExecutor<Resource, ResourceCriteria> queryExecutor,
+                                                       PropertyOptionsSource.ExpressionScope expressionScope,
+                                                       Predicate<Resource> baseResourcePredicate) {
+        Iterable<Resource> foundResources;
+
+        criteria.setSearchExpression(expression);
+        foundResources = new CriteriaQuery<Resource, ResourceCriteria>(criteria, queryExecutor);
```
burmanm: Merge the definition here.

josejulio: Could you elaborate a bit more?

burmanm: Well, there's no difference to this code:

```
criteria.setSearchExpression(expression);
Iterable<Resource> foundResources = new CriteriaQuery<Resource, ResourceCriteria>(criteria, queryExecutor);
```

Now the separate definition causes some confusion if that foundResources could be loaded from elsewhere also, which isn't true (I'm assuming you've done refactoring here which is why it looks like this).

josejulio: Done, thanks!

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
josejulio: Done, thanks!

josejulio:
@loleary, @spinder: 
I'm not sure how to do that, not sure if is possible with the current code.

As is, I inject the additional `base resource type` into the expression (e.g. `expression="regular_query;base_type"`). 
It works if server is updated.
if it isn't it breaks the expression parser. It doesn't know how to process that.

I tried looking to see if there was any way to insert `comments` on the expression (defined by [this](https://github.com/rhq-project/rhq/blob/56cab9e579247325fc815bbfbc123e40f6c95d7a/modules/enterprise/server/jar/src/main/antlr3/org/rhq/enterprise/server/search/RHQL.g#L25-L25) grammar) but couldn't find any way.

The only options I can think of are: 
- "hard-code" on server side that validation instead of using the rhq-plugin.xml.
- Provide that configuration in a separate file. That way it would degrade well as that new file wouldn't be used by old versions or the server and only new versions would know how to load (if it exists) to fetch additional configurations.

Thoughts?

//cc @burmanm @rubenvp8510

loleary:
> "hard-code" on server side that validation instead of using the rhq-plugin.xml.

How would this get around the dependency issue? If the validation was on the server side, it would be able to handle the case of an old plug-in or new plug-in. But, if they had not yet applied the server patch, the new plug-in would break the agent/server.

I normally would be against hard-coding the validation but as this is a fix for a downstream maintenance release, it might be the best option. If the plug-in code itself could be updated to use the appropriate expression depending on the version of the server, that would be great. But I am not sure this is possible. Keep in mind that servers are updated separate from agents and plug-ins. Plug-ins are updated separate from agents and servers. Agents are updated separate from plug-ins and servers. Meaning that a user's environment could be made up of agent v1 *and* v2, along with server v1 *or* v2, along with plug-in v1 *and* v2.

> Provide that configuration in a separate file. That way it would degrade well as that new file wouldn't be used by old versions or the server and only new versions would know how to load (if it exists) to fetch additional configurations.

Definitely don't want a separate configuration file. 

Going back to your original plan of modifying the XSD. Is it possible to do this and add the new logic as an optional attribute? In other words, if `rhq-plugin.xml` includes the new attribute and the server has the old XSD, would it fail? Or do we allow unknown attributes to be ignored when validating?

josejulio:
> loleary commented 8 minutes ago
> > "hard-code" on server side that validation instead of using the rhq-plugin.xml.
>
> How would this get around the dependency issue? If the validation was on the server side, it would be able to handle the case of an old plug-in or new plug-in. But, if they had not yet applied the server patch, the new plug-in would break the agent/server.

The server would check if:
1. If using option-source with expression-scope=BASE_RESOURCE
2. Plugin  == EAP6 OR EAP7.
3. The current resource has a `Profile` type on any of its ancestors.

If this meets, i would follow from here: https://github.com/rhq-project/rhq/pull/303/files#diff-7899cd03a510cfc8a0ec452471e4a31bR2669
with ancestorType = 'Profile'

> I normally would be against hard-coding the validation but as this is a fix for a downstream maintenance release, it might be the best option. If the plug-in code itself could be updated to use the appropriate expression depending on the version of the server, that would be great. But I am not sure this is possible. Keep in mind that servers are updated separate from agents and plug-ins. Plug-ins are updated separate from agents and servers. Agents are updated separate from plug-ins and servers. Meaning that a user's environment could be made up of agent v1 and v2, along with server v1 or v2, along with plug-in v1 and  v2.

This issue only affects the Configuration UI, so we won't require any plug-in update.The server code that builds the configuration UI uses the rhq-plugin.xml to decide what values to show.

> > Provide that configuration in a separate file. That way it would degrade well as that new file wouldn't be used by old versions or the server and only new versions would know how to load (if it exists) to fetch additional configurations.
>
> Definitely don't want a separate configuration file.

Got it.

> Going back to your original plan of modifying the XSD. Is it possible to do this and add the new logic as an optional attribute? In other words, if rhq-plugin.xml includes the new attribute and the server has the old XSD, would it fail? Or do we allow unknown attributes to be ignored when validating?

It would fail, I tried it.

The old server would complain with the new plugin (with new attributes) something like 'Attribute xyz is not expected..."

josejulio:
I guess the best would be to hardcode this validation without requiring any plug-in update. I'll work on a PoC and open a new PR.

burmanm:
How about adding an expression that will always match - but that we can detect to be something else?

Here's an example, that will always work in the older version of JON (because there's a bug in the CriteriaQueryGenerator):

In rhq-plugin.xml:

```xml
<c:option-source target="resource" expression="type=SocketBindingGroup profile.plugin=&pluginName;" expressionScope="baseResource"/>
```

This is going to CriteriaQueryGenerator and there's a transformtoJPQL (line 759), which gets:

```
searchExpression = type=SocketBindingGroup profile.plugin=JBossAS7
translatedJPQL = (  ( LOWER(resource.resourceType.name) LIKE '%socketbindinggroup%' )  AND  ( LOWER(resource.resourceType.plugin) LIKE '%jbossas7%' )  ) 
```

You can put anything infront of the actual column name (the ``plugin.`` part) and it will be filtered out nicely in the current JON version. Now, you could read this info instead of ``;`` and do the parsing.

Works in both versions, old style in the old JONs and new style in the new JON versions.

burmanm:
I originally tried to get the parentResource exploited to get the information you wanted (instead of the Java code), but I ran into that bug so I couldn't actually get the parentResource to do my bidding.

josejulio:
@burmanm Awesome!

That might work well for my needs, thanks!

josejulio:
I still need to test this, but the idea is there and should work.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
```diff
@@ -2655,17 +2655,24 @@ private void handlePDS(final Subject subject, Resource resource, Resource parent
 
             Predicate<Resource> baseResourcePredicate = null;
             if (expressionScope == PropertyOptionsSource.ExpressionScope.BASE_RESOURCE) {
-                int halfCommaIndex = expression.indexOf(";");
-                if (halfCommaIndex > -1) {
+                String[] expressionTokens = expression.split("\\\\s+");
+                String ancestorType = null;
+                final Pattern ancestorTypePattern = Pattern.compile("(.*)\\.plugin");
```
burmanm: Just in case we want to exploit this behavior for some other feature, should we have a more exact match than just anything with ``.plugin`` for this ancestor search?

Like ``.ancestor.plugin`` or something like that with a prefix being the one that's being searched. Namespacing bugs is obviously quite funny, but ..

josejulio: Sure, I'll use `.base_resource.plugin`.

josejulio: while `Profile.plugin=EAP7` yields

```sql
 ( LOWER(resource.resourceType.plugin) LIKE '%eap7%' ) 
```

`profile.ancestor.plugin` yields
```sql
 ( LOWER(resource.name) LIKE '%profile.ancestor.plugin=eap7%' ) 
```

Luckily `Profile;baseresource.plugin=EAP7` yields
```sql
 ( LOWER(resource.resourceType.plugin) LIKE '%eap7%' ) 
```

josejulio: I'm going to use `;` to separate the value from the type,
I guess that using any of [these](https://github.com/josejulio/rhq/blob/56cab9e579247325fc815bbfbc123e40f6c95d7a/modules/enterprise/server/jar/src/main/antlr3/org/rhq/enterprise/server/search/RHQL.g#L174) will break this behavior.
e.g. `Profile\\\\;baseresource.plugin=EAP7` still yields
```sql
 ( LOWER(resource.resourceType.plugin) LIKE '%eap7%' ) 
```

@burmanm What do you think we use `value;name.plugin=EAP7` to inject extra arguments? value can be optional and left blank.

josejulio:
@burmanm could you please give another review round?

burmanm:
It seems I get the notifications little late, but is the ";" version sure to work without your patch?

josejulio:
It should, my patch doesn't (after my last commit) modify the expression itself, so the ; goes to the expression query. See samples here: https://github.com/rhq-project/rhq/pull/303#discussion_r113009988

josejulio:
Rebased and did some changes to wfly-10 and jboss-as-7 plugin xml. 
Still need to test them.

Some changes are to use baseResource and avoid mixing servers resources on a RHQ instance. Others are to restrict the resources to a Profile.

modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -6055,7 +6055,7 @@
               <c:simple-property name="connectors" />
             </c:list-property>
             <c:simple-property name="socket-binding" required="true" type="string" readOnly="false" description="The broadcast group socket binding.">
-              <c:option-source target="configuration" expression="*/socket-binding=name:type=SocketBindingGroup"/>
+              <c:option-source target="configuration" expression="*/socket-binding=name:type=SocketBindingGroup" expressionScope="baseResource"/>
```
josejulio: @burmanm I added `expressionScope="baseResource"` to all `c:option-source` elements. All of them seems to expect its own resources anyway. Do you think this is too aggressive?

josejulio:
> It seems I get the notifications little late, but is the ";" version sure to work without your patch?

It works, but we get this bad looking line on the logs:
```
13:37:31,870 ERROR [stderr] (http-/0.0.0.0:7080-3) line 1:7 no viable alternative at character ';'
```

I suppose is the parser complaining...

josejulio:
I was trying to make this a bit generic so we could put any other type in case we needed that. 
Maybe that won't be needed at all, since if we try to use any *funny* character like `\`, `;`, `½`, `¬` we get that not so funny error  `line 1:7 no viable alternative at character ';'` which might cause confusion.
In the end I'm going to go for
ProfileBaseResource.plugin=EAP7. So if we would like to use another type for a base resource it would be like OtherTypeBaseResource.plugin=EAP7...

josejulio:
This is ready to be merged.
Thanks to @burmanm the core and plugin won't crash/fail if both aren't on the last version. However for this to be fixed we still require both plugin and core to be up to date.

PR #306, opened 2017-05-24
[BZ-1448500] Remove/ignore unused properties for servlet on profile=l…

josejulio:
…oad-balancer

  Ignore setting=jsp/websocket when loading data.
  Remove setting=jsp/websocket when updating data.
  Add "(Not used by load-balancer)" on display name for setting=jsp/websocket.

@burmanm could you take a look?

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ConfigurationLoadDelegate.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ConfigurationLoadDelegate.java
@@ -183,11 +183,12 @@ public class ConfigurationLoadDelegate implements ConfigurationFacet {
         Result operationResult = connection.execute(op);
         if (!operationResult.isSuccess()) {
             // check if the definition is needed -- required==true or not. Only complain for required ones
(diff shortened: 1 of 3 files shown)


burmanm:
Does this still work with EAP7 also?

josejulio:
It should! Let me test.

josejulio:
It does work.
I'm querying the resource to know if it has the setting=jsp and setting=websocket properties, if they are found everything goes as usual. 
The ones that aren't found are removed (the whole group, e.g. all setting=jsp/*)

loleary:
Why not leave the label intact and just use a special value to indicate its not used?

For example, on configuration read, if the property is not supported (i.e. load-balancer profile) then the value returned could be "*** not used by load-balancer ***". Even if the user changes the value and saves it, it would be ignored and the control would have its value updated to "*** not used by load-balancer ***".

This prevents hard-coding the control's display name to include a profile that could later change or be expanded.

Regardless, the approach you are taking seems to be the right one form my point-of-view.

josejulio:
@loleary I like that idea. I recall that most fields were booleans, but will double check and see the idea fits well there.

josejulio:
@burmanm I did @loleary suggestions.
Could you please review?

PR #307, opened 2017-06-07
Bug 1436568 - Missing "Max Connections" metrics for http listener in …

rubenvp8510:
…EAP 7

### DIFF
--- a/modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
+++ b/modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
@@ -3142,6 +3142,7 @@
           <metric property="max-processing-time" units="nanoseconds" description="The maximum processing time taken by a request on this listener"/>
           <metric property="processing-time" units="nanoseconds" measurementType="trendsup" description="The total processing time of all requests handed by this listener"/>
           <metric property="request-count" measurementType="trendsup" description="The number of requests this listener has served"/>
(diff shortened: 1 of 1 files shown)


modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -3142,6 +3142,7 @@
           <metric property="max-processing-time" units="nanoseconds" description="The maximum processing time taken by a request on this listener"/>
           <metric property="processing-time" units="nanoseconds" measurementType="trendsup" description="The total processing time of all requests handed by this listener"/>
           <metric property="request-count" measurementType="trendsup" description="The number of requests this listener has served"/>
+          <metric property="_expr:max-connections" displayName="Max connections" description="Number of max connections this connector can handle (see configuration to set it)" defaultOn="true" defaultInterval="14400000"/>
```
josejulio: What does the `_expr:` do?

rubenvp8510:
@josejulio It is for solve an expression (1) if the attribute has one, you can see how it works on (2) and (3).

1) https://docs.jboss.org/author/display/AS71/Expressions
2) https://github.com/rhq-project/rhq/blob/master/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/BaseComponent.java#L213

3) https://github.com/rhq-project/rhq/blob/master/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/BaseComponent.java#L291

PR #308, opened 2017-07-18
Bug 1469292 - Now RPC ping support NFSv3 

rubenvp8510:


modules/core/native-system/src/main/java/org/rhq/core/system/FileSystemInfo.java
```diff
@@ -68,12 +71,39 @@ private void createFileSystemObject(SigarProxy sigar) {
 
     }
 
+    private boolean nfsPing(String hostname) {
+        return  (RPC.ping(hostname, RPC.TCP, NFS_PROGRAM, 2) == 0) ||
```
loleary: Perhaps the first attempt should be with version 3 as it is probably the most common now? Might even be worth adding version 4 as a final attempt? So, TCP v3, UDP v3, TCP v2, UDP v2, TCP v4, UDP v4?

rubenvp8510:
This PR is already updated with support for NFSv4, also did changes in the ping order, according to @loleary  comments.

PR #310, opened 2017-08-02
[BZ-1401565] Validate if the server has already started before call R…

rubenvp8510:
…EST handlers.

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/SetCallerInterceptor.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/SetCallerInterceptor.java
@@ -82,11 +82,10 @@ public class SetCallerInterceptor {
         Subject caller=null;
         java.security.Principal p = ejbContext.getCallerPrincipal();
         if (!startupBean.isInitialized()) {
-            String notInitMessage = "Tried to call REST endpoint but the server has not finished to startup";
-            log.warn(notInitMessage);
-            Response.ResponseBuilder builder;
-            builder = Response.status(Response.Status.SERVICE_UNAVAILABLE).entity(notInitMessage);
-            return builder.build();
+            String notInitMessage = "Tried to call REST endpoint but the server is not ready - still booting up";
+            log.debug(notInitMessage);
+            return Response.status(Response.Status.SERVICE_UNAVAILABLE)
+                    .header("Retry-After", "30").entity(notInitMessage).build();
         }
 
         if (p!=null) {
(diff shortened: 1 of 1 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/SetCallerInterceptor.java
```diff
@@ -72,6 +81,14 @@ public Object setCaller(InvocationContext ctx) throws Exception {
 
         Subject caller=null;
         java.security.Principal p = ejbContext.getCallerPrincipal();
+        if (!startupBean.isInitialized()) {
+            String notInitMessage = "Tried to call REST endpoint but the server has not finished to startup";
+            log.warn(notInitMessage);
```
loleary: Do we even want to log this? If we feel this message may be useful, perhaps it should only get logged at DEBUG level? Just seems that "expected" failures such as "I'm not running yet" don't warrant a log message in the log of the thing that "is not running yet."

Log messages should be considered valid in exceptional or informational state change events. "I am starting..." "I have started." "I am not going to start due to failure..." etc.

If there is no way to tell that the server had not started from the context of the surrounding log messages then perhaps a DEBUG message would be helpful in those rare cases that someone cannot figure out that the REST service is unavailable because the server is not running.

rubenvp8510: I changed the message to debug, Is not that this is not running yet or not running at all, but it's in the initialization process, may be loading some database configurations, but the REST API is already deployed and the EAP server is already running so the message is logged property (at least on my tests).

In this particular case, happens that `getDefaultDatabaseType` attribute is null because isn't set yet when someone tries to consume the REST service.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/SetCallerInterceptor.java
```diff
@@ -72,6 +81,14 @@ public Object setCaller(InvocationContext ctx) throws Exception {
 
         Subject caller=null;
         java.security.Principal p = ejbContext.getCallerPrincipal();
+        if (!startupBean.isInitialized()) {
+            String notInitMessage = "Tried to call REST endpoint but the server has not finished to startup";
+            log.debug(notInitMessage);
+            Response.ResponseBuilder builder;
```
burmanm: Combine lines 87 and 88.

On another note, should we set a retry header here like we do in the AuthenticationFilter on legacy UI?

rubenvp8510:
@burmanm I've already combined the lines, also I think it's a good idea to set the header.

PR #311, opened 2017-08-02
Bug 1416097 - Failed to delete the bundles when deleting a bundle wit…

josejulio:
…h lot of configuration parameters and versions

 Added an INFO message telling it takes long to delete a bundle with many parameters and versions.


modules/enterprise/gui/coregui/src/main/resources/org/rhq/coregui/client/Messages.properties
```diff
@@ -1571,6 +1571,8 @@ view_bundle_list_deleteConfirm = Are you sure you want to delete the selected bu
 view_bundle_list_deleteFailure = Failed to delete the bundle [{0}]
 view_bundle_list_deleteSubmitted = A request to delete the selected bundles has been submitted.
 view_bundle_list_deleteSuccessful = You successfully deleted the bundle [{0}]
+view_bundle_list_deleteTakesLong = A request to delete the selected bundles is taking too long
+view_bundle_list_deleteTakesLongDetail = Server is still processing deletion of bundles. Bundles with many parameters and versions are know to take a long time.
```
loleary: My opinion... a better message might be:

`The server is still processing the deletion of one or more bundles. Bundles with many parameters or versions can take several minutes to delete. The bundle will continue to appear in the bundle list view until the deletion is complete and the view is refreshed.`

josejulio: Thanks! That is a way better message!

PR #315, opened 2017-08-17
Force  updateBackingContent to discover the new package in sync manner.

rubenvp8510:
Not tested yet, but this should fix BZ-1420712 and BZ-1420717

Don't merge yet.

### DIFF
--- a/modules/enterprise/binding/src/main/java/org/rhq/bindings/client/ResourceClientProxy.java
+++ b/modules/enterprise/binding/src/main/java/org/rhq/bindings/client/ResourceClientProxy.java
@@ -46,8 +46,10 @@ import org.rhq.core.domain.configuration.PluginConfigurationUpdate;
 import org.rhq.core.domain.configuration.ResourceConfigurationUpdate;
 import org.rhq.core.domain.configuration.definition.ConfigurationDefinition;
 import org.rhq.core.domain.content.InstalledPackage;
+import org.rhq.core.domain.content.PackageDetailsKey;
 import org.rhq.core.domain.content.PackageType;
 import org.rhq.core.domain.content.PackageVersion;
+import org.rhq.core.domain.content.transfer.ResourcePackageDetails;
 import org.rhq.core.domain.criteria.MeasurementDefinitionCriteria;
 import org.rhq.core.domain.criteria.MeasurementScheduleCriteria;
 import org.rhq.core.domain.criteria.OperationDefinitionCriteria;
@@ -616,6 +618,13 @@ public class ResourceClientProxy {
 
             contentManager.deployPackagesWithNote(remoteClient.getSubject(), new int[] { resourceClientProxy.getId() },
                 new int[] { pv.getId() }, "CLI deployment request");
+            PackageDetailsKey keys = new PackageDetailsKey(
+                    oldPackage.getPackageVersion().getGeneralPackage().getName(),
+                    packageVersion,
+                    oldPackage.getPackageVersion().getGeneralPackage().getPackageType().getName(),
+                    oldPackage.getPackageVersion().getArchitecture().getName());
+            ResourcePackageDetails details = new ResourcePackageDetails(keys);
+            contentManager.mergePackage(remoteClient.getSubject(), resourceClientProxy.getId(), details);
         }
 
         public void retrieveBackingContent(String fileName) throws IOException {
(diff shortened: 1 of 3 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/content/ContentManagerRemote.java
```diff
@@ -284,11 +283,13 @@ PackageTypeAndVersionFormatComposite findPackageTypeWithVersionFormat(Subject su
...
-     *
      * @since 4.10
      */
     PackageVersion createPackageVersionWithDisplayVersion(Subject subject, String packageName, int packageTypeId,
-        String version, String displayVersion, Integer architectureId, String temporaryContentHandle);
+                                                          String version, String displayVersion, Integer architectureId,
+                                                          String temporaryContentHandle);
+
+    //To force to discovery certain package
+    void mergePackage(int resourceId, ResourcePackageDetails details);
```
burmanm: Why is this exposed in the Remote?

rubenvp8510: Because it's used by `ResourceClientProxy` and it needs this exposed as a remote method

loleary: Then that will be a problem. All remote calls must be authenticated. 

Also, it does seem very strange that this would be exposed as a remote call because on its own, it has no real purpose? Is there a better way to do this? Methods that go into the Remote interface should be those that will be invoked directly by a remote client.

burmanm:
@rubenvp8510 Although the commit made comments outdated, the issue itself has not been solved

rubenvp8510:
Added subject and permission validations to the new remote method.

josejulio:
It looks like its already authenticated.

loleary:
How do you mean it is already authenticated? What am I missing?

jshaughn:
I could be way off the mark, but if I remember, ResourceClientProxy adds the auth implicitly given the CLI login creds, or something like that.  maybe that is what Ruben is referring to, I'm not sure.

rubenvp8510:
I created a new remote method, at the beginning I forgot to pass subject and validate for permissions, but I've already updated the PR with that. And yes, from the `ResourceClientProxy` I can get the subject there and pass it to the remote method.

modules/enterprise/binding/src/main/java/org/rhq/bindings/client/ResourceClientProxy.java
```diff
@@ -616,6 +618,13 @@ public void updateBackingContent(String filename, String displayVersion) {
 
             contentManager.deployPackagesWithNote(remoteClient.getSubject(), new int[] { resourceClientProxy.getId() },
                 new int[] { pv.getId() }, "CLI deployment request");
+            PackageDetailsKey keys = new PackageDetailsKey(
+                    oldPackage.getPackageVersion().getGeneralPackage().getName(),
+                    packageVersion,
+                    oldPackage.getPackageVersion().getGeneralPackage().getPackageType().getName(),
+                    oldPackage.getPackageVersion().getArchitecture().getName());
+            ResourcePackageDetails details = new ResourcePackageDetails(keys);
+            contentManager.mergePackage(remoteClient.getSubject(), resourceClientProxy.getId(), details);
```
burmanm: Is this really blocking? I mean, what happens if the agent takes ages to actually deploy the package? Are we now returning backing content that's not really actually deployed necessarily yet?

rubenvp8510:
I did another commit because sometimes the installed package id changes and the old one doesn't exist anymore so `getPackageBytes` method cannot retrieve the bits (because the package doesn't exist anymore). This is probably due some other job runs in an async way, (probably the purge job or when the deploy process finished)  and change or replace the installed package.

In the last commit, I tried to retrieve the backingContent (installedPackage) again and ask for the bits again. (I set a timeout on this..).

 In the tests I did this happens sometimes. not too much, and if this happens the retrieveBackingContent takes time to get the bits.

I deployed packages with 10Mb > and update the package with 150+ versions,  this seems to not fail.

josejulio:
I tested and seems to be working OK. (No exceptions)

burmanm:
Didn't this fail because of the missing intentional-changes.xml ?

josejulio:
> Didn't this fail because of the missing intentional-changes.xml ?

I didn't run those tests. But yes, @rubenvp8510 needs to add/update the relevant intentional-changes.xml

josejulio:
LGTM, pending to add the relevant intentional-changes.xml file.

PR #316, opened 2017-08-18
Bug 1420060 - Sleep when server is up but agent still needs to register

josejulio:
We already know that the server is up, but the agent haven't registered yet.
This should give some time to the agent to complete its registration process.

### DIFF
--- a/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java
+++ b/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java
@@ -718,6 +718,7 @@ public class AgentMain {
                     boolean mustRegister = prepareStartupWorkRequiringServer();
                     boolean keepWaitingForServer;
                     int serverUpWaitCounter;
+                    long serverUpSleepMillis = 10000L;
                     try {
                         // we really should not need to have to customize this, but make a backdoor setting, just in case
                         serverUpWaitCounter = Integer.parseInt(System.getProperty(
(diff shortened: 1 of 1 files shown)


josejulio:
@burmanm can you please review?

I couldn't reproduce the reported issue, but i'm guessing that the server is available at some point, but registration thread is [sleeping](https://github.com/josejulio/rhq/blob/6899a0dcb8d1383b6ee227452408b856b21d37c9/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java#L1540), since server is available it doesn't wait and exits.

burmanm:
Based on @FilipB 's comments, this guessed behavior is not the real reason (some agents were able to register)

josejulio:
This is my only guess so far (as couldn't reproduce). 
If you look at the bottom of the agent log[1], you can see that `[RHQ Agent Registration Thread]` could indeed talk to the server for registration (the server rejected because of token), but the main thread is already finishing the agent.

```
2017-02-07 11:19:25,398 ERROR [RHQ Agent Registration Thread] (org.rhq.enterprise.agent.AgentMain)- {AgentMain.agent-registration-rejected}The server has rejected the agent registration request. Cause: [org.rhq.core.clientapi.server.core.AgentRegistrationException:The agent [fbr-ag-rpm6.bc.jonqe.lab.eng.bos.redhat.com] is attempting to re-register without a security token. Please consult an administrator to obtain the agent's proper security token and restart the agent with the option "-Drhq.agent.security-token=<the valid security token>". An administrator can find the agent's security token by navigating to the GUI page "Administration (Topology) > Agents" and drilling down to this specific agent. You will see the long security token string there. For more information, read: https://docs.jboss.org/author/display/RHQ/Agent+Registration]
```
My guess is: Server is busy (not sure reason), when it gets unbusy,  `[RHQ Agent Registration Thread]`  is [sleeping](https://github.com/josejulio/rhq/blob/6899a0dcb8d1383b6ee227452408b856b21d37c9/modules/enterprise/agent/src/main/java/org/rhq/enterprise/agent/AgentMain.java#L1540) but main thread already noticed agent is up (as you can see, it doesn't wait) and completes the 5 rounds of test almost immediately, agent is not registered, we can connect to server, shut it down.


[1] https://bugzilla.redhat.com/attachment.cgi?id=1248471

PR #317, opened 2017-08-22
Bug 1380471 - Check for timeouts when getting live availability

josejulio:
- Set availability to unknown on timeout


modules/enterprise/gui/coregui/src/main/resources/org/rhq/coregui/client/Messages.properties
```diff
@@ -2011,6 +2011,7 @@ view_inventory_mixed = mixed
 view_inventory_platforms = Platforms
 view_inventory_problemGroups = Problem Groups
 view_inventory_resource_loadFailed = Resource with id [{0}] does not exist or is not accessible
+view_inventory_resource_liveAvailabilityTimeout = Unable to obtain this resource current availability state.
```
burmanm: How about only "Unable to obtain resource's availability state" ? (or just availability without the state word)

josejulio: `Unable to obtain resource's availability` works for me.

josejulio: Updated the text

PR #319, opened 2017-08-30
Bug 1447965 - Scale chart units using average instead of lowest value.

rubenvp8510:


josejulio:
LGTM

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/common/graph/MetricGraphData.java
```diff
@@ -365,21 +365,18 @@ public String getJsonMetrics() {
...
-                    }
-                    if (measurement.getLowValue() < lowestValue.getLowValue()) {
-                        lowestValue = measurement;
-                    }
+                    averageData += measurement.getValue();
                 }
             }
-            if (null != lowestValue) {
+
+            averageData /= metricData.size();
```
burmanm: This is tricky. Are we supposed to calculate average from the amount of metrics given that they were not all necessarily valid?

For example, given a series of: [NaN, NaN, 5.0, NaN, NaN] the average is 1.0. But should it actually be 5.0 ? The NaN values are not metrics - we don't know what's in them. I don't think they should be calculated to the average. They're not in the graph either.

rubenvp8510: I saw that the average is already calculated, used that value instead.

PR #321, opened 2017-09-06
Bug 1417937 - Updates columnsWidth to match the column count

josejulio:
Firefox rendered an extra column because of this.


rubenvp8510:
LGTM

PR #322, opened 2017-09-11
Bug 1429909 - pass group id to alert definition delete method.

rubenvp8510:


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/alert/AlertDefinitionManagerLocal.java
```diff
@@ -164,4 +164,18 @@ AlertDefinition updateAlertDefinitionInternal(Subject subject, int alertDefiniti
...
+     * <p/>
+     * Requires MANAGE_ALERTS permission for the group or resource on which the alert definition is defined. REQUIRES
+     * MANAGE_SETTINGS for Templates.
+     *
+     * @param subject
+     * @param groupId
+     * @param alertDefinitionIds
+     * @return number of removed definitions
+     */
+    int removeAlertDefinitions(Subject subject,Integer groupId, int[] alertDefinitionIds);
```
josejulio: I suppose it would make sense to expose this as a public API, as is the only way to remove alert definitions using group permissions.

loleary:
It seems to me that the proper fix for this would be to just remove the child alert definitions. If the specified rhq_alert_definition.id has a null resource then a list of alert definition ids can be queried by retrieving all rows from rhq_alert_definition with group_alert_def_id = id.

I do not see why a new method is needed. Please help me understand.

Thanks.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/alert/AlertDefinitionManagerBean.java
```diff
@@ -343,6 +348,10 @@ public int removeAlertDefinitions(Subject subject, int[] alertDefinitionIds) {
         criteria.addFilterIds(ArrayUtils.wrapInArray(alertDefinitionIds));
         criteria.addFilterDeleted(false);
         criteria.clearPaging();
+
+        if ( groupId != null ){
+            criteria.addFilterResourceGroupIds(groupId);
```
loleary: Does this actually work? From what I gather, it does not change anything. When I select an alert definition from a group's alert definition page the alert id is the group alert definition. The `groupId` is not relevant.

rubenvp8510: It supposes to not be relevant, but the thing is how the AlertDefinitionCriteria works, according to this: 

https://github.com/rubenvp8510/rhq/blob/8d1ede47c80394e9bc7a587d21c1ed6b3baf9417/modules/core/domain/src/main/java/org/rhq/core/domain/criteria/AlertDefinitionCriteria.java#L37-L37

I need to set the ```groupId``` in order to get the AlertDefinitions,  if I don't do it the query doesn't return anything.  This is why for the cases on the users that don't have Inventory Manager permissions but have Alert Definition permissions, the alert is not deleted, because of this:

https://github.com/rubenvp8510/rhq/blob/03438c5b843693a26f29d0c853d47eea4b8b7635/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/alert/AlertDefinitionManagerBean.java#L355-L355

Returns 0.

rubenvp8510: I could modify this, but I would prefer to not do it.

loleary: So will the delete cascade then? I think that was what my concern was. Thanks for the clarification.

rubenvp8510: The groupId is just for permission validations, as the comment on AlertDefinitionCriteria said.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/alert/AlertDefinitionManagerLocal.java
josejulio: I suppose we could keep this method variation under private API unless we are requested to expose it

//cc @spinder @loleary

loleary: Although I would like it to be Public, I think you might be right about keeping it Private until someone explicitly requests it. Reasoning being:
 - JBoss ON is in maintenance mode.
 - This is really an enhancement.
 - The workaround of finding all resources that use the alert definition and manually removing it from said resources is doable via the API.
 - To expose it in the Public API, QE would need to define test cases and update test plans to include it for their remote API test suites.

josejulio:
LGTM. Reproduced the bug and this PR fixes it.

PR #324, opened 2017-09-12
Bug 1455091 - Attemp to fix non valid values when starting the plugin

josejulio:
This checks everytime a `ServerPlugin` is initialized if we have a valid value for the transport if not, writes the default configuration onto the database.
// cc @loleary

### DIFF
--- a/modules/enterprise/server/plugins/alert-snmp/src/main/java/org/rhq/enterprise/server/plugins/alertSnmp/SNMPComponent.java
+++ b/modules/enterprise/server/plugins/alert-snmp/src/main/java/org/rhq/enterprise/server/plugins/alertSnmp/SNMPComponent.java
@@ -1,0 +1,63 @@
+package org.rhq.enterprise.server.plugins.alertSnmp;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


modules/enterprise/server/plugins/alert-snmp/src/main/java/org/rhq/enterprise/server/plugins/alertSnmp/SNMPComponent.java
```diff
@@ -0,0 +1,63 @@
...
+
+                String transportValue = configuration.getSimpleValue(PROP_TRANSPORT, null);
+                if (transportValue == null || !VALID_TRANSPORTS.contains(transportValue)) {
+                    ServerPluginManagerLocal spm = LookupUtil.getServerPluginManager();
+                    ServerPlugin serverPlugin  = spm.getServerPlugin(context.getPluginEnvironment().getPluginKey());
+                    serverPlugin.getPluginConfiguration().setSimpleValue(PROP_TRANSPORT, DEFAULT_TRANSPORT);
+                    LookupUtil.getServerPluginManager().updateServerPluginExceptContent(
+                            LookupUtil.getSubjectManager().getOverlord(),
+                            serverPlugin
+                    );
```
josejulio: @burmanm are you aware of other way to do this? (Save the configuration for a `ServerPlugin`)
// cc: @rubenvp8510

PR #328, opened 2017-09-21
[Bug 1447965] Use MetricDisplaySummary fit units for chart scaling.

rubenvp8510:
@josejulio  @burmanm  Could you please review?

### DIFF
--- a/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/monitoring/table/MetricsTableView.java
+++ b/modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/monitoring/table/MetricsTableView.java
@@ -21,6 +21,9 @@ package org.rhq.coregui.client.inventory.resource.detail.monitoring.table;
 
 import static org.rhq.coregui.client.inventory.resource.detail.monitoring.table.MetricsGridFieldName.METRIC_DEF_ID;
 import static org.rhq.coregui.client.inventory.resource.detail.monitoring.table.MetricsGridFieldName.RESOURCE_ID;
+import static org.rhq.coregui.client.inventory.resource.detail.monitoring.table.MetricsGridFieldName.DISPLAY_UNITS_NAME;
+import static org.rhq.coregui.client.inventory.resource.detail.monitoring.table.MetricsGridFieldName
+        .DISPLAY_UNITS_FAMILY;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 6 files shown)


modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/monitoring/table/MetricsTableView.java
```diff
@@ -361,6 +365,18 @@ public void expandOpenedRows() {
...
+            MeasurementUnits recordUnits = null;
+            if (record.getAttributeAsString(DISPLAY_UNITS_NAME.getValue())!=null && record.getAttributeAsString
+                    (DISPLAY_UNITS_FAMILY.getValue())!=null) {
+
+                recordUnits = MeasurementUnits.getUsingDisplayUnits(
+                        record.getAttributeAsString(DISPLAY_UNITS_NAME.getValue()),
+                        MeasurementUnits.Family.valueOf(record.getAttributeAsString(DISPLAY_UNITS_FAMILY.getValue())));
+            }
+
+            final MeasurementUnits recordDisplayUnits = recordUnits;
```
burmanm: I can't see really from this github UI, but what's the point of this assignment?

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/util/MeasurementConverterClient.java
```diff
@@ -141,6 +141,21 @@ public static String getMeasurementUnitAbbreviation(MeasurementUnits units) {
         }
     }
 
+    public static MeasurementUnits bestFitUnits(double[] values,  MeasurementUnits originalUnits) {
+        // find bestFit units by taking the average
+        Double average = 0.0;
+        for (int i = 0, sz = values.length; i < sz; i++) {
```
burmanm: Why not just use values.length instead of ``sz`` ? To me at least ``values.length`` is more readable

rubenvp8510: +1, I just copied the source code and made it a function, I'll change.

modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/monitoring/table/MetricsTableView.java
rubenvp8510: This is because I need to make if final to access it from the inner AsyncCallback anonymous object.

https://github.com/rhq-project/rhq/pull/328/files/c1a6281b8a0f33132df33a4dff6d71fc4bd79436#diff-07cbfda325e06a09752feca7adc9bbebR417

PR #329, opened 2017-09-22
Bug 1455091 - Uses configuration being sent instead of fetching

josejulio:
I was getting `could not initialize proxy - no Session` when trying to load the configuration from the ServerPlugin. That issue wasn't showing on RHQ.
I build this plugin on RHQ and installed it on JON, and was working as expected with this change.


josejulio:
@rubenvp8510 or @burmanm 
Please review

PR #330, opened 2017-10-03
JWS-820 - Fail to load Tomcat8 datasource configuration

maxime-beck:


rhqci:

PR #331, opened 2017-10-19
BZ 1477945 Added environment and cache fields to subsystem=naming/binding entries

rubenvp8510:
Still need some test.

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingBindingComponent.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingBindingComponent.java
@@ -1,0 +1,107 @@
+package org.rhq.modules.plugins.wildfly10;
+
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import org.rhq.core.domain.configuration.Configuration;
+import org.rhq.core.domain.configuration.definition.ConfigurationDefinition;
+import org.rhq.modules.plugins.wildfly10.json.Operation;
+import org.rhq.modules.plugins.wildfly10.json.Result;
+import org.rhq.modules.plugins.wildfly10.json.ReadAttribute;
+import org.rhq.core.pluginapi.configuration.ConfigurationUpdateReport;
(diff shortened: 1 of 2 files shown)


rubenvp8510:
@josejulio Could you please review.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingBindingComponent.java
```diff
@@ -0,0 +1,107 @@
...
+        ConfigurationDefinition configDef = context.getResourceType().getResourceConfigurationDefinition().copy();
+        /* Get and remove enviroment property */
+        configDef.getPropertyDefinitions().remove(ENVIRONMENT);
+
+        /* Load other properties*/
+        ConfigurationLoadDelegate delegate = new ConfigurationLoadDelegate(configDef, getASConnection(), address,
+                includeRuntime);
+        Configuration configuration = delegate.loadResourceConfiguration();
+
+        Operation op = new Operation("whoami", getAddress());
```
burmanm: Why? Why not just check the result of the real operation if it fails or not?

rubenvp8510: I just take it from existing code :) but yes, we can check just for the result, I can change it.

josejulio: Not sure if this is what you meant:

```java
        //:whoami might fail host controller resources, in that case use :read-resource operation
        //which is slower due to larger content returned but more reliable since every resource has it.

```
https://github.com/rhq-project/rhq/blob/master/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/BaseComponent.java#L369-L370

rubenvp8510: As I saw this is just for know if the server needs to be restarted or  need to be reloaded after configuration changes, so the "real" operation here is `whoami`, which is used just for check for those two things (reload and restart flags), the `resource-reading` op is a fallback in cases where `whoami` fails.

PR #332, opened 2017-10-24
JWS-834 - Apache plugin fails to retrieve mod_bmx metrics when listen is set to *:port

maxime-beck:


rhqci:

modules/plugins/apache/src/main/java/org/rhq/plugins/apache/ApacheServerDiscoveryComponent.java
```diff
@@ -370,6 +370,10 @@ private DiscoveredResourceDetails discoverSingleProcess(
             /* try mod_bmx VirtualHost and Location */
             String bmxURL = findBMXURL(serverConfig);
             if (bmxURL != null) {
+                String[] split_url = bmxURL.split(":");
+                if(split_url[1].equals("//*"))
```
burmanm: No chance for ArrayIndexOutOfBoundsException? Is the input checked somewhere that it always includes all the parts?

burmanm: .. appears to be.

PR #333, opened 2017-10-31
Bug 1477945 Environment and cache parameters added to subsystem=naming/binding fo…

rubenvp8510:
…r EAP6

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/resources/META-INF/rhq-plugin.xml
+++ b/modules/plugins/jboss-as-7/src/main/resources/META-INF/rhq-plugin.xml
@@ -2184,7 +2184,7 @@
       </operation>
 
       <service name="Binding (Managed Server)"
-               class="BaseComponent"
+               class="NamingBindingComponent"
                discovery="SubsystemDiscovery"
                description="JNDI bindings for primitive types">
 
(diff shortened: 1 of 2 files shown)


modules/plugins/jboss-as-7/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -5922,6 +5929,13 @@
           <c:simple-property name="module" required="false" type="string" readOnly="false" description="The module to load the object factory from for object factory bindings"/>
           <c:simple-property name="type" required="false" type="string" readOnly="false" description="The type of the value to bind for simple bindings, this must be a primitive type"/>
           <c:simple-property name="value" required="false" type="string" readOnly="false" description="The value to bind for simple bindings"/>
+          <c:simple-property name="cache" required="false" type="boolean" readOnly="false" defaultValue="false" description="If the external context should be cached"/>
+          <c:list-property name="environment" readOnly="true" description="A list of The environment to use on object factory instance retrieval." required="false">
+            <c:map-property name="environment" readOnly="true" displayName="Environment">
```
josejulio: why `readOnly=true` on the list and the map?

rubenvp8510: My mistake, should be readonly= false,

rubenvp8510:
Updated with the observations, @josejulio could you do another quick review, please? Thanks

josejulio:
LGTM

PR #336, opened 2017-11-01
Bug 1503826 - Change redeploy check, using deploy name instead of runtime-name

rubenvp8510:
@josejulio Could you please review?

//cc @burmanm

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/Redeployer.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/Redeployer.java
@@ -35,14 +35,16 @@ import org.rhq.modules.plugins.jbossas7.json.Result;
  */
 class Redeployer {
 
+    private final String name;
     private final String runtimeName;
     private final String hash;
     private final ASConnection connection;
 
-    Redeployer(String runtimeName, String hash, ASConnection connection) {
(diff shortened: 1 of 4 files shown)


josejulio:
Sorry, forgot to fill in this before.

With this fix and after following the steps on the BZ I get a different runtime-name than the one specified (the filename).

I'm not sure if this is related to this fix (possible) or the Bundle script used to test.
con-el-fix

rubenvp8510:
@josejulio  I fixed that issue, this was happening because the redeploy operation didn't set any runtime-name, so by default the runtime-name is the filename.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/Redeployer.java
```diff
@@ -59,7 +61,10 @@ boolean deploymentExists() {
 
     Result redeployOnServer() {
         Operation op = new Operation("full-replace-deployment", new Address());
-        op.addAdditionalProperty("name", runtimeName);
+        op.addAdditionalProperty("name", name);
+        if (runtimeName != null) {
+            op.addAdditionalProperty("runtime-name", runtimeName);
```
rubenvp8510: @josejulio I added the runtime-parameter here, if it is specified (not-null) I'll pass as a parameter of the full-replace-deployment op.

josejulio:
LGTM

PR #337, opened 2017-11-02
Bug 1504148 - Change redeploy check, using deploy name instead of runtime-name

rubenvp8510:
This is the same as BZ 1503826  but this is for EAP7

@burmanm @josejulio


josejulio:
LGTM

PR #340, opened 2018-01-10
Bug 1418034 - Fix orpahaned configuration when update config parameters.

rubenvp8510:
@josejulio Could you please review,

This only mitigates the problem, there is still some orphaned configurations that appear from time to time, I'm not sure where that comes from.

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
@@ -345,9 +345,6 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
         resource.addPluginConfigurationUpdates(update);
         entityManager.remove(resource.getPluginConfiguration());
         resource.setPluginConfiguration(update.getConfiguration().deepCopyWithoutProxies());
-
-        entityManager.merge(update);
-
         return update;
     }
 
@@ -378,7 +375,6 @@ public class ConfigurationManagerBean implements ConfigurationManagerLocal, Conf
         }
         entityManager.remove(resource.getResourceConfiguration());
         resource.setResourceConfiguration(configuration);
-        entityManager.merge(resource);
     }
 
     // Use new transaction because this only works if the resource in question has not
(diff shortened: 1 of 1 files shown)


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
```diff
@@ -237,6 +238,7 @@ public void completePluginConfigurationUpdate(PluginConfigurationUpdate update)
             // Flush before merging to ensure the update has been persisted and avoid StaleStateExceptions.
             entityManager.flush();
             entityManager.merge(update);
+            entityManager.remove(oldPluginConfig);
```
loleary: Wouldn't we want the old plug-in configuration to remain in the plug-in config history page of the resource? It seems that a configuration update/merge would not result in an orphaned configuration object. Granted, if the configuration object has no associated configuration parameters, I would consider the update to be strange/broken.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/configuration/ConfigurationManagerBean.java
```diff
@@ -1568,7 +1570,9 @@ public void completeResourceConfigurationUpdate(ConfigurationUpdateResponse resp
         } else if (response.getStatus() == ConfigurationUpdateStatus.SUCCESS) {
             // link to the newer, persisted configuration object
             Resource resource = update.getResource();
+            Configuration oldResourceConfig = resource.getResourceConfiguration();
             resource.setResourceConfiguration(update.getConfiguration().deepCopyWithoutProxies());
+            entityManager.remove(oldResourceConfig);
```
loleary: Same question/concern as before. Wouldn't updated resource configuration simply result in the previous configuration history to appear on the resource's configuration history page?

jshaughn:
LGTM

josejulio:
I'm sure that this isn't something we want to do. This would result in deleting the previous configuration (which is not orpahaned AFAIK). 
Old configurations are displayed on the history, but with this change we always delete it.

screenshot from 2018-01-10 12-23-44

Maybe i'm missing something?

rubenvp8510:
mm I'll review this in more detail, it seems like just in some situations the orphaned happens. but it happens for sure because I tested it.

jshaughn:
For what it's worth, although there may be some simple explanation for some leaked config, it's more likely something very subtle in the entity mgmt.  More in line with the fixes done for BZ 1323332.  Possibly something to do with cascade delete.

jshaughn:
@rubenvp8510 OK, these changes look good to me.  Adding the copy for the pluginConfig saved the history, right?   I think we just need to add the one other case we talked about.   For reference, other reviewers may want to see the comments in https://github.com/rubenvp8510/rhq/pull/1.

PR #341, opened 2018-01-11
Bug 1532387 - Create child alerts for groups and templates

josejulio:
@burmanm can you please review?

### DIFF
--- a/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/AlertDefinitionHandlerBean.java
+++ b/modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/rest/AlertDefinitionHandlerBean.java
@@ -87,6 +87,8 @@ import org.rhq.enterprise.server.alert.AlertConditionManagerLocal;
 import org.rhq.enterprise.server.alert.AlertDefinitionManagerLocal;
 import org.rhq.enterprise.server.alert.AlertManagerLocal;
 import org.rhq.enterprise.server.alert.AlertNotificationManagerLocal;
+import org.rhq.enterprise.server.alert.AlertTemplateManagerLocal;
+import org.rhq.enterprise.server.alert.GroupAlertDefinitionManagerLocal;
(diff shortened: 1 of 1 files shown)


josejulio:
Updated to also create it for templates and removed the `InNewTransaction` code

jshaughn:
I haven't looked at detail at what you're doing but certainly if you want to support group and template alert defs via the remote API it needs to call into the correct underlying services.  What caught my eye was just, "removed the InNewTransaction code".   I just want to mention that these 'InNewTransaction' services are typically there for a reason, so be very careful in any restructuring.  They are there sometimes to prevent large umbrella transactions and sometimes to ensure committed data before performing higher level work.  You may want to look at the git history to see why it's there.  I apologize in advance if you've already done that investigation, I'm just being overly cautious as many times the trasactioning work has been done to solve prior issues.

josejulio:
Jay, Thanks for your comments.
What i meant by "removed the InNewTransaction code" is:
When starting to work on this BZ, I saw that this code was being called:
```alertDefinitionManager.createAlertDefinitionInNewTransaction```

But the `GroupDefinitionManager` and `AlertTemplateManager` don't have a similar code, at the beginning I created a method `xxxInNewTransaction` for GroupDefinitionManager because I wasn't sure if there is any special reason to have that. In the end removed that method because, `GroupDefitinionManager` calls under the hood ```alertDefinitionManager.createAlertDefinitionInNewTransaction```.

jshaughn:
@josejulio Ah, OK, yes that makes sense.  The group/template level code typically will break the work into smaller transactions by doing exactly that, calling an 'xxxInNewTransaction' on each group member or type instance, for example.  In this way we avoid the umbrella transactions that cause timeouts or massive resource use, and commit work incrementally.  So yes, you don't want that 'inNewTransaction' type logic at the group or template level.  Thanks for the response.  We use the 'InNewTransaction' naming to easily reflect that the service/function has the non-default transaction attribute and so that callers more easily know what they are doing.

josejulio:
@burmanm could you do a review round? is ready for merge

PR #342, opened 2018-01-17
Bug 1532749, 1532743 - Keep message parent div always on top (z-index)

josejulio:
Instead of resetting to auto all the z-index which are managed by gwt
Give the message div the highest (32bit max signed int) priority

@rubenvp8510 please review


rubenvp8510:
LGTM

PR #346, opened 2018-03-07
Bug 1537720, 1538697 - Adds default value for isLocal

josejulio:
Also Fix tests related to manually adding a server

It was failing because two things:

- We didn't mock the `getServerPluginConfiguration` method, it wasn't being used on these tests.
- We didn't provide a default value in case "isLocal" is `null`, this could happen when importing using an old version and then upgrading, defaults to False, old versions always used the "Remote" path.

### DIFF
--- a/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/ServerPluginConfiguration.java
+++ b/modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/ServerPluginConfiguration.java
@@ -182,7 +182,7 @@ public class ServerPluginConfiguration {
     }
 
     public Boolean isLocal() {
-        return this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
+        return Boolean.valueOf(this.pluginConfig.getSimpleValue(Property.LOCAL, "false"));
     }
 
     public void setLocal(boolean isLocal) {
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/helper/ServerPluginConfiguration.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/helper/ServerPluginConfiguration.java
@@ -160,7 +160,8 @@ public class ServerPluginConfiguration {
     }
 
     public Boolean isLocal() {
-        return this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
+        return Boolean.valueOf(this.pluginConfig.getSimpleValue(Property.LOCAL, "false"));
+
     }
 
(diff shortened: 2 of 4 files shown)


josejulio:
@burmanm please take a look

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/ServerPluginConfiguration.java
```diff
@@ -182,7 +182,11 @@ public Long getManagementConnectionTimeout() {
     }
 
     public Boolean isLocal() {
-        return this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
+        Boolean isLocal = this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
```
burmanm: This can cause NPE. ``this.pluginConfig.getSimple(Property.LOCAL)`` is equal to ``null`` as defaultValue is property is not found.

burmanm: So my recommendation here is:

```java
return Boolean.valueOf(this.pluginConfig.getSimpleValue(Property.LOCAL, "false"));
```

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/helper/ServerPluginConfiguration.java
```diff
@@ -160,7 +160,11 @@ public Long getManagementConnectionTimeout() {
     }
 
     public Boolean isLocal() {
-        return this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
+        Boolean isLocal = this.pluginConfig.getSimple(Property.LOCAL).getBooleanValue();
```
burmanm: Same here as above.

modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/helper/ServerPluginConfiguration.java
josejulio: Whoops, right on that

josejulio: Updated code

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/helper/ServerPluginConfiguration.java
josejulio: Updated code

PR #348, opened 2018-08-17
Bug 1542125 - Allows rest user to logout to be able to refresh its permissions

josejulio:
- Set SecurityDomain[RHQRESTSecurityDomain] cache-type to infinispan
 - Creates infinispan cache at server install
   (named security, which is what cache-type=infinispan uses)
 - Creates new endpoint /user/logout to force the removal of the
   logged user cache

Reverts https://github.com/rhq-project/rhq/commit/5837457f16bdd16d7c956bfbeb549de0fe478e13 (i.e. reverts the revert)

This make the following changes to `standalone-full.xml` while installing:

- security-domain name="RHQRESTSecurityDomain" cache-type is changed from `default` to `infinispan`
- The following `cache-container` is created:
```xml
<cache-container name="security" default-cache="auth-cache" jndi-name="java:jboss/infinispan/security">
    <local-cache name="auth-cache">
        <eviction strategy="LRU" max-entries="1000"/>
        <expiration max-idle="200000" lifespan="900000"/>
     </local-cache>
</cache-container>
```

### DIFF
--- a/modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
+++ b/modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
@@ -406,7 +406,11 @@ public class ServerInstallUtil {
             DELEGATING_LOGIN_MODULE_NAME, AppConfigurationEntry.LoginModuleControlFlag.SUFFICIENT, options);
 
         SecurityDomainJBossASClient client = new SecurityDomainJBossASClient(mcc);
-        client.createNewSecurityDomain(RHQ_REST_SECURITY_DOMAIN, loginModuleRequest);
+        InfinispanJBossASClient infinispanClient = new InfinispanJBossASClient(mcc);
+
+        // Create the cache used for the rest security domain
+        infinispanClient.createSecurityDomainInfinispanCache("LRU", 1000, 200000L, 900000L);
+        client.createNewSecurityDomain(RHQ_REST_SECURITY_DOMAIN, true, loginModuleRequest);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


loleary:
I am assuming that the `java:jboss/infinispan/security` _cache-container_ can be added using the apply-updates command for existing installations? 

It also seems that 5837457f16bdd16d7c956bfbeb549de0fe478e13 was doing other stuff. Was it not? 

From the BZ I can see that you tested but do not see the test scenario in where the role was removed and you waited for the cache expiration to ensure that all was good.

josejulio:
> I am assuming that the java:jboss/infinispan/security cache-container can be added using the apply-updates command for existing installations?

I think it can, but isn't in this PR (yet) I will work on that.

> It also seems that 5837457 was doing other stuff. Was it not?

From what I understood, is still needed, that code was executed once in the life of the server (non expiring cache of 1000 users) and ensures we are picking the latest groups of the user, my changes ensures that the cache expires and the code in 5837457 is executed again

> From the BZ I can see that you tested but do not see the test scenario in where the role was removed and you waited for the cache expiration to ensure that all was good.

I think I tested that but didn't document enough. I'll do another test round and be more specific in the bz

josejulio:
apply-updates code won't be here, so this is ready for review

modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
```diff
@@ -406,7 +406,11 @@ public static void createRestSecurityDomain(ModelControllerClient mcc) throws Ex
             DELEGATING_LOGIN_MODULE_NAME, AppConfigurationEntry.LoginModuleControlFlag.SUFFICIENT, options);
 
         SecurityDomainJBossASClient client = new SecurityDomainJBossASClient(mcc);
-        client.createNewSecurityDomain(RHQ_REST_SECURITY_DOMAIN, loginModuleRequest);
+        InfinispanJBossASClient infinispanClient = new InfinispanJBossASClient(mcc);
+
+        // Create the cache used for the rest security domain
+        infinispanClient.createSecurityDomainInfinispanCache("LRU", 1000, 200000L, 900000L);
```
burmanm: Is LRU really the correct format?

josejulio:
I think this one is ready for merge.

PR #349, opened 2018-08-22
Bug-1517717 Allow to specify rpc_address and listen_address in rhq-st…

josejulio:
…orage.properties

Still need to somehow test this works.


rubenvp8510:
I need to test this, but LGTM.

josejulio:
I did some tests here to check if the values were applied correctly:
1) only hostname
2) hostname + rpc_address
3) hostname + listen_address
4) hostname + rpc_address + listen_address

It seems that is picking the right values, only concern I have is that i'm not sure if should use listen_address for the seeds, but you can always specify the seeds.

rubenvp8510:
@josejulio Is correct, for seeds you should use the listen address if IRC is the address/port used by gossip protocol and the node-cluster communication.

RCP is just for clients.

PR #351, opened 2018-09-20
Bug 1542125 - Allow the server to start even if we don't have the req…

josejulio:
…uired cache-container for the rest logout

 This allows the updater to start the server to properly configure it.


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/util/LookupUtil.java
```diff
@@ -282,6 +283,19 @@ public static EntityManager getEntityManager() {
         }
     }
 
+    public static CacheContainer getSecurityCacheContainer() {
+        try {
+            InitialContext context = new InitialContext();
+            CacheContainer cacheContainer = (CacheContainer) context.lookup("java:jboss/infinispan/security");
+            context.close();
```
burmanm: In case of exception, this resource stays open. This should be in the } finally { part (or you could use the existing methods such as lookup which do this already)

PR #352, opened 2018-09-21
Bug 1542125 - Ensure we always close the context

josejulio:


modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/util/LookupUtil.java
```diff
@@ -285,10 +285,7 @@ public static EntityManager getEntityManager() {
 
     public static CacheContainer getSecurityCacheContainer() {
         try {
-            InitialContext context = new InitialContext();
-            CacheContainer cacheContainer = (CacheContainer) context.lookup("java:jboss/infinispan/security");
-            context.close();
-            return cacheContainer;
+            return (CacheContainer) lookup("java:jboss/infinispan/security");
```
josejulio: @burmanm I suppose you meant this lookup method.

PR #355, opened 2018-10-02
Bug 1630170 - Set the same properties as jboss-cli-commands.txt

josejulio:
- Allows the installer to write jboss.tx.node.id even if not set at install time
 - Sets jackson.deserialization.whitelist.packages

### DIFF
--- a/modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
+++ b/modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
@@ -241,11 +241,11 @@ public class ServerInstallUtil {
         client.setDefaultTransactionTimeout(600);
         LOG.info("Default transaction timeout set to 600 seconds.");
 
-        if(serverProperties.containsKey(ServerProperties.PROP_JBOSS_HA_NODE_ID)) {
-            String nodeId = buildExpression(ServerProperties.PROP_JBOSS_HA_NODE_ID, serverProperties, true, false, false);
-            client.setTransactionNodeId(nodeId);
-            LOG.info("Set high availability nodeId to " + nodeId);
-        }
+        // Set expression and its default regardless if a value was specified at install time
+        final HashMap<String, String> transactionNodeIdDefault = new HashMap();
+        transactionNodeIdDefault.put(ServerProperties.PROP_JBOSS_HA_NODE_ID, "1");
+        String nodeId = buildExpression(ServerProperties.PROP_JBOSS_HA_NODE_ID, transactionNodeIdDefault, true, false, false);
+        client.setTransactionNodeId(nodeId);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
```diff
@@ -241,11 +241,12 @@ public static void configureTransactionManager(ModelControllerClient mcc, HashMa
...
-            String nodeId = buildExpression(ServerProperties.PROP_JBOSS_HA_NODE_ID, serverProperties, true, false, false);
-            client.setTransactionNodeId(nodeId);
-            LOG.info("Set high availability nodeId to " + nodeId);
-        }
+        // Set expression and its default regardless if a value was specified at install time
+        final HashMap<String, String> transactionNodeIdDefault = new HashMap();
+        transactionNodeIdDefault.put(ServerProperties.PROP_JBOSS_HA_NODE_ID, "1");
+        String nodeId = buildExpression(ServerProperties.PROP_JBOSS_HA_NODE_ID, transactionNodeIdDefault, true, false, false);
+        client.setTransactionNodeId(nodeId);
+        LOG.info("Set high availability nodeId to " + nodeId);
```
burmanm: I think this logging should be removed if we're not setting anything that's been defined. It could be confusing otherwise-

modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
```diff
@@ -261,6 +262,11 @@ public static void configureDeploymentScanner(ModelControllerClient mcc) throws
         LOG.info("Deployment scanner turned off.");
     }
 
+    public static void configureSystemProperties(ModelControllerClient mcc) throws Exception {
+        CoreJBossASClient client = new CoreJBossASClient(mcc);
+        client.setSystemProperty("jackson.deserialization.whitelist.packages", "org.rhq.enterprise.server.rest.domain");
```
burmanm: This is technically only something that works in the JON-release (master branch has different jackson version). Did you test with the master that it does not break anything?

josejulio: I didn't check the logs. I installed and started RHQ and checked that everything was in there.
Will double check that nothing seems broken.

modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
josejulio: +1

modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
josejulio: Tested and it doesn't seem to break anything on master branch.

modules/enterprise/server/installer/src/main/java/org/rhq/enterprise/server/installer/ServerInstallUtil.java
josejulio: Removed.

josejulio:
@burmanm please review again

PR #358, opened 2018-12-10
Bug 1654629 - Raise certificate timeout from 10s to 30s

josejulio:


josejulio:
I'll try something else

PR #359, opened 2018-12-12
Bug 1654629 Allow to configure the detect-certificate-timeout

josejulio:
To configure, we have to pass to the rhq-agent:
  -Drhq.agent.plugin.wildfly10.detect-certificate-timeout=20
 Where 20 is the number of seconds to wait before timing out.

 This value can be specified inside the rhq-agent-env.sh as part
 of the RHQ_AGENT_JAVA_OPTS

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ServerControl.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/ServerControl.java
@@ -216,13 +216,23 @@ public final class ServerControl {
          * whether to execute dummy CLI check prior running desired command
          */
         private boolean checkCertificate = true;
+        private long detectCertificateTimeout = 10 * 1000L;
 
 
         Cli() {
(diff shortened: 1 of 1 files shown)


josejulio:
An alternative for #358

josejulio:
//cc: @loleary @spinder

burmanm:
A very large number for the timeout makes no sense (there's something else wrong in the env) and having a larger default should not be an issue for normal behavior as the timeout is never met. For those reasons I would say #358  was enough.

loleary:
I agree with @burmanm. Adding a configuration option is not necessary. We already have so many "timeout" related configuration options I would like to either use one of those or just use what you did in #358. Is it safe to assume that a meaningful timeout message is produced that makes it clear that it took longer then the allowed 30 seconds? Something that makes it clear why it fails and that something is potentially wrong in the environment?

josejulio:
> Is it safe to assume that a meaningful timeout message is produced that makes it clear that it took longer then the allowed 30 seconds? Something that makes it clear why it fails and that something is potentially wrong in the environment?

The timeout is used to know if the user needs to accept the certificate, if the operation takes long that the expected timeout, we assume that jboss-cli is stuck on :
`Accept certificate? [N]o, [T]emporarily, [P]ermanently: `
If so, we raise an error telling the user that is needed to manually accept the certificate.

PR #361, opened 2019-02-06
Patches smartgwt-skin (aka SmartClient) file to disallow the use of

josejulio:
-moz-scrollbar on firefox 63+

I'm patching the external dependency (smartgwt-skin)
Is obfuscated, but since is unlikely that we will change it from 3.0, i think is OK as long as we document the patch.

I tested with chrome, firefox 24 and firefox 65.

PR #362, opened 2019-02-25
Patches smartgwt to have the scrollbars working in firefox 63+

josejulio:
This is intended for jon3.3.x, similar to #361 . The only difference is that the code is applied to line 2100 instead of 2065


burmanm:
@spinder This is more for you, are you happy with this maven change in the jon3.3.x release branch?

PR #364, opened 2019-03-05
Bug 1636056 - Fix the loading image name

josejulio:


modules/enterprise/gui/coregui/src/main/java/org/rhq/coregui/client/inventory/resource/detail/ResourceTitleBar.java
```diff
@@ -81,7 +81,7 @@
     private static final String COLLAPSED_TOOLTIP = MSG.view_portlet_inventory_tooltip_expand();
     private static final String EXPANDED_TOOLTIP = MSG.view_portlet_inventory_tooltip_collapse();
     private static final String PLUGIN_ERRORS_ICON = "subsystems/inventory/Resource_warn.png";
-    //private static final String LOADING_ICON = "[SKIN]/loading.gif";
+    //private static final String LOADING_ICON = "[SKIN]/loadingSmall.gif";
```
burmanm: Isn't this kinda useless patch? ;) Just remove the line.

josejulio: Ah, I left it as a reference, but i can remove it.

josejulio: Done

PR #366, opened 2019-03-11
Bug 1636061 - Ignore groups with null cn

josejulio:


josejulio:
Didn't saw any other place where the LDAP groups were fetched.

josejulio:
@burmanm could you check?

burmanm:
As far as I know, that's the only way to do it. LDAP itself does not support (in every implementation) a filter for return values and this is the only file where we do the queries.

modules/enterprise/server/jar/src/main/java/org/rhq/enterprise/server/resource/group/LdapGroupManagerBean.java
```diff
@@ -638,8 +638,11 @@ private void executeGroupSearch(String filter, Set<Map<String, String>> groupDet
             }
             //
             Map<String, String> entry = new HashMap<String, String>();
-            String name = (String) si.getAttributes().get("cn").get();
-            name = name.trim();
+            Attribute nameAttribute = si.getAttributes().get("cn");
+            if (nameAttribute == null) {
+                continue;
+            }
+            String name = ((String) nameAttribute.get()).trim();
```
burmanm: Just a small question.. can the nameAttribute be non-null but the value itself from ``.get()`` be null? Or should it be checked in the above ``nameAttribute == null || nameAttribute.get() == null`` ?

josejulio: I'm not sure. 
I would think that if we already have an value it will be non null.
I'll add the extra test to be sure of that.

PR #370, opened 2019-03-28
Bug 1655950 - Makes "Io Threads" a required attribute

josejulio:


modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -3348,7 +3348,7 @@
             </plugin-configuration>
 
             <resource-configuration>
-                <c:simple-property name="io-threads" required="false" type="integer" readOnly="true" description="Specify the number of I/O threads to create for the worker.  If not specified, a default will be chosen, which is calculated by cpuCount * 2"/>
+                <c:simple-property name="io-threads" required="true" type="integer" readOnly="true" description="Specify the number of I/O threads to create for the worker.  If not specified, a default will be chosen, which is calculated by cpuCount * 2"/>
```
burmanm: If it's required.. how do we calculate the cpuCount * 2 ?

josejulio: There is maybe a bug in EAP, if you try to set this value to undefined, EAP will throw an exception. I think this will be solved differently.

PR #371, opened 2019-03-28
Bug 1655950 - Sets a dummy value to  get into reload-required status as required

josejulio:
This allows to safely set undefined to io-threads and task-max-threads, other way we would
  get a IllegalArgumentException

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/WorkerComponent.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/WorkerComponent.java
@@ -1,3 +1,22 @@
+/*
+ * RHQ Management Platform
+ * Copyright (C) 2005-2014 Red Hat, Inc.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


josejulio:
@loleary  @burmanm can you comment on this approach?

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/WorkerComponent.java
```diff
@@ -0,0 +1,33 @@
+package org.rhq.modules.plugins.wildfly10;
```
loleary: Missing copyright/license header.

loleary:
I think this approach is dangerous but it does seem to be a valid workaround.

My reasoning is that we are applying the workaround blindly -- Regardless of version or what configuration attributes are actually changing. That said, it does seem that all the configuration attributes for an io worker resource require reload. Meaning that if any configuration changes are made, reload would be required. Therefore, this does make sense to me. I am curious to what @burmanm thinks.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/WorkerComponent.java
josejulio: Thanks

burmanm:
@loleary We talked about this in the IRC, a short recap here. For other values (like the task-keepalive mentioned in the ticket) we should do the update without this trick and instead just remove these values which we can't update from the batch update sent to the EAP. For these values which we can't currently unset in the EAP, we would first contact the EAP team and see if it's a bug that should be fixed in the actual product. And we'll wait a bit for their reply before merging any workarounds.

josejulio:
Closing in favor of #377

PR #372, opened 2019-03-29
Bug 1652063 - Overrides creation for NamingComponent for support environment propertly

rubenvp8510:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1652063


modules/plugins/jboss-as-7/src/main/java/org/rhq/modules/plugins/jbossas7/NamingComponent.java
```diff
@@ -1,11 +1,11 @@
 package org.rhq.modules.plugins.jbossas7;
 
-import java.util.Map;
+import java.util.*;
```
josejulio: Do we really need it all?

rubenvp8510: Haa noo of course no 😅

PR #374, opened 2019-04-02
Bug 1652061 - Overrides creation for NamingComponent for support envi…

rubenvp8510:
…ronment key value pair list

### DIFF
--- a/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingComponent.java
+++ b/modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingComponent.java
@@ -1,11 +1,21 @@
 package org.rhq.modules.plugins.wildfly10;
 
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.rhq.core.domain.configuration.Configuration;
+import org.rhq.core.domain.configuration.Property;
 import org.rhq.core.domain.configuration.PropertyList;
 import org.rhq.core.domain.configuration.PropertyMap;
 import org.rhq.core.domain.configuration.PropertySimple;
+import org.rhq.core.domain.configuration.definition.ConfigurationDefinition;
+import org.rhq.core.domain.configuration.definition.PropertyDefinition;
+import org.rhq.core.domain.configuration.definition.PropertyDefinitionList;
+import org.rhq.core.domain.configuration.definition.PropertyDefinitionMap;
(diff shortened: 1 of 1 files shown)


rubenvp8510:
@josejulio

josejulio:
Code looks good to me

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingComponent.java
```diff
@@ -17,6 +27,48 @@
...
+            /* Modify resource configuration on-fly*/
+            PropertyMap environmentMap = new PropertyMap();
+            environmentMap.setName(ENVIRONMENT);
+            Map<String, Property> results = new HashMap<String, Property>();
+            for (Property inner : embeddedProps) {
+                PropertyMap propMap = (PropertyMap) inner;
+                PropertySimple propName = ((PropertySimple) propMap.get("name"));
+                PropertySimple propValue = ((PropertySimple) propMap.get("value"));
+                PropertySimple propS = new PropertySimple();
+                propS.setName(propName.getStringValue());
```
burmanm: Are these null safe?

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingComponent.java
```diff
@@ -17,6 +27,48 @@
  * @author Heiko W. Rupp
  */
 public class NamingComponent extends BaseComponent implements OperationFacet {
+    private static final String BINDING_TYPE = "Binding";
+    private static final String ENVIRONMENT = "environment";
+    @Override
+    public CreateResourceReport createResource(CreateResourceReport report) {
+        if (report.getResourceType().getName().equals(BINDING_TYPE)) {
+            Configuration resourceConfiguration = report.getResourceConfiguration();
+            PropertyList prop = (PropertyList)resourceConfiguration.get(ENVIRONMENT);
```
burmanm: Actually, same to all of these ``.get()`` are all of these always present with the ``BINDING_TYPE`` or can something be null also? There's no null checks in this part of the code.

modules/plugins/wfly-10/src/main/java/org/rhq/modules/plugins/wildfly10/NamingComponent.java
rubenvp8510: Good point, I think I need to add null validations here to avoid NPE.

rubenvp8510:
Added some null validations! Thanks for the review!

PR #377, opened 2019-04-08
Bug 1655950 - Sets a dummy value to  get into reload-required status …

josejulio:
…as required

  This allows to safely set undefined to io-threads and task-max-threads, other way we would
  get a IllegalArgumentException


josejulio:
This will only write the dummy value as needed.

@loleary @burmanm

PR #378, opened 2019-04-09
Bug 1655647 - Fix path for jaspi test

josejulio:


modules/plugins/wfly-10/src/main/resources/META-INF/rhq-plugin.xml
```diff
@@ -13031,11 +13031,11 @@
 
 	&loginModuleResourceConfig;
 
-    <service name="Login Module Stack" discovery="ModuleOptionsDiscoveryComponent" class="ModuleOptionsComponent"
```
loleary: Why are we removing Login Module Stack? This will break any existing installs.

josejulio: This configuration is broken, I assumed EAP6 did things differently but it seems i was wrong (seems broken to me as well). The obvious choice was to use the other configuration which was more in line with the ModuleOptionsComponent

